PR #4, opened 2019-12-23
Document Optional and Default params

StephenBrown2:
Optional params will always [follow the pattern](https://docs.python.org/3/library/typing.html#typing.Optional) `Union[X, NoneType]`, so we can match on it and give a more succinct type string.

Also, I saw the vistigial code for default params, so I figured I'd work it back in, using `*required*` as a stand-in for params without a default value.


pawamoy:
Ha, good. Yeah I previously removed the default column because I didn't think it had real value since we can see the defaults in the signature itself. Also the parameters names or types sometimes get wrapped on two lines, so I wanted to give them more space by removing the default column.

I'm gonna merge this since I guess it has value to you, and we'll make it configurable later :slightly_smiling_face:

PR #7, opened 2019-12-26
wip: Use extension

pawamoy:
Just a reminder of how we will do this. Not to be merged.

The roadmap is:
1. refactor documenter.py to really separate the build process from the rendering process
2. move the rendering code to an extension (from "markdown to markdown+features" to "markdown to html")

### DIFF
--- a/src/mkdocstrings/extension.py
+++ b/src/mkdocstrings/extension.py
@@ -1,0 +1,51 @@
+from markdown import Markdown
+from markdown.extensions import Extension
+from markdown.blockprocessors import BlockProcessor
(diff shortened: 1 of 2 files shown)


StephenBrown2:
Not sure if it's currently possible with mkdocs at all, but adding a watcher to the source code to regen the docs when docstrings are modified would be a fantastic addition.

pawamoy:
Indeed. I think I saw something about watchers in the doc.

pawamoy:
Closing in favor of #41 (wip)

Issue #27, opened 2020-01-07
Support setting argument/return types in the docstring


thebigmunch:
Not everybody uses (or wants to use) type hints in their code. Not supporting setting types in the docstrings limits the user base of mkdocstrings to those that do.

Would you be amenable to adding support for setting the types in the docstring? It would probably make most sense to follow the [Google style](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google) as done with sections.

Thanks for the consideration and the promising package.

pawamoy:
I don't really want to add support for types in docstrings. Here is why:

- as usual, added complexity, work, maintenance work. If someone is willing to be responsible for its support, then why not, but I myself do not volunteer 🙂 
- I consider that, if you need comprehensive documentation of your Python code, you probably need type annotations anyway, to ensure quality of your code. It's totally subjective of course, and I have no problem with people thinking otherwise
- I think type annotations is a good addition to ensure quality of a project, and thus am trying to promote it. You know, being opinionated.
- Google-style sections are valuable to me because, contrary to `:param:` and the likes, they help keeping the docstring readable in the Python interpreter, with the `help` builtin. But types in these same docstrings are now kinda replaced, in a better way, by type annotations (well, I guess having the type written next to the param name helps readability as well, so this point might not stand...)

**However**. I know type annotations are still not widely used. It's starting, but *a lot* of projects are still not using them. As you correctly said, not supporting types in docstrings limits the user base of `mkdocstrings` (though I'm not targetting every project. mkdocstrings is particularly focused on modern/recent projects, for which it's easy to add type annotations).

Therefore, maybe we *could* add support for types in docstrings, *for those who are transitioning to type annotations*, but still, I am not sure about the value of that.

Lets leave that issue open and wait for feedback / upvotes.

P.S.: for those who wonder, types hints in comments (`# type: int`) are a big *no*.

thebigmunch:
Type annotations actively decrease readability of code. For people that are familiar with them, they **might** increase the understandability of the code. But they are mainly for static analysis tools. I, for one, will never use them in my hobby code. I write that code specifically to be as accessible to novices as possible. Type annotations very much go in the opposite direction.

While you might think it is best for projects to have them, there are many good reasons Python core devs responsible for them said they shouldn't be required as far as Python itself is concerned.

[...]

PR #37, opened 2020-01-26
handle types in docstring args, Exception types

cs01:
Fix 1:
The signature cannot be determined from this code.
```
class MyError(ValueError):
    """Raise when trying to ..."""

    pass
```

 This PR uses an empty signature when signature cannot be determined. 

Fix 2:
When a docstring for arguments has a type hint, an error would be thrown. This PR strips out text in parentheses from the argument name, and allows documentation to be successfully generated.

(Unfortunately I do not have the time to write full tests for these fixes, but wanted to share them anyway)

### DIFF
--- a/src/mkdocstrings/docstrings.py
+++ b/src/mkdocstrings/docstrings.py
@@ -212,7 +212,14 @@ class Docstring:
         block, i = self.read_block_items(lines, start_index)
         for param_line in block:
             try:
-                name, description = param_line.lstrip(" ").split(":", 1)
+                name_with_type, description = param_line.lstrip(" ").split(":", 1)
+                paren_index = name_with_type.find("(")
+                if paren_index != -1:
+                    # name (type)
+                    name = name_with_type[0:paren_index].strip()
+                else:
+                    # no type, just use name as-is
+                    name = name_with_type
             except Exception as e:
                 print(f"Failed to get 'name: description' pair from '{param_line}'")
                 continue
--- a/src/mkdocstrings/docstrings.py
+++ b/src/mkdocstrings/docstrings.py
@@ -211,7 +211,11 @@ class Docstring:
         parameters = []
         block, i = self.read_block_items(lines, start_index)
         for param_line in block:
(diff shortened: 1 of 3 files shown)


src/mkdocstrings/documenter.py
```diff
@@ -342,12 +342,13 @@ def get_class_documentation(self, class_: Type[Any], module: Optional[ModuleType
...
-            path=path,
-            file_path=file_path,
-            docstring=Docstring(textwrap.dedent(class_.__doc__ or ""), inspect.signature(class_)),
-        )
+        try:
+            signature = inspect.signature(class_)
+        except ValueError:
+            print(f"Failed to get signature for {class_name}")
+            signature = inspect.Signature()
+        docstring = Docstring(textwrap.dedent(class_.__doc__ or ""),)
```
pawamoy: I think you forgot to pass the `signature` to the `Docstring` constructor 😛 ?

src/mkdocstrings/docstrings.py
```diff
@@ -211,7 +211,18 @@ def read_parameters_section(self, lines, start_index):
         parameters = []
         block, i = self.read_block_items(lines, start_index)
         for param_line in block:
-            name, description = param_line.lstrip(" ").split(":", 1)
+            try:
+                name_with_type, description = param_line.lstrip(" ").split(":", 1)
+                paren_index = name_with_type.find("(")
+                if paren_index != -1:
+                    # name (type)
+                    name = name_with_type[0:paren_index].strip()
```
pawamoy: I like that, it could help implementing #27

cs01: Definitely, would be easy to add a regex to extract the type if it were present.

src/mkdocstrings/documenter.py
cs01: Thanks, fixed

pawamoy:
Alright, there's a conflict to resolve and we're good to go!

pawamoy:
Thank you @cs01!

PR #38, opened 2020-01-26
do not render empty types

### LINKED ISSUE #34: Replace `_empty` with `` in the `Type` column
Right now if a type annotation is not present, it's rendered as `_empty`. It should just be an empty table cell if there is no type, otherwise it looks like it's using a type called `_empty`. It could also say "unknown" or "not defined".

cs01:
If there is no type associated with an argument, display an empty table cell rather than the word `_empty`.

fixes #34

### DIFF
--- a/src/mkdocstrings/docstrings.py
+++ b/src/mkdocstrings/docstrings.py
@@ -44,7 +44,6 @@ TITLES_RETURN = ("return:", "returns:")
 
 RE_OPTIONAL = re.compile(r"Union\[(.+), NoneType\]")
 RE_FORWARD_REF = re.compile(r"_ForwardRef\('([^']+)'\)")
-RE_EMPTY = re.compile(r"_empty")
 
 
 class AnnotatedObject:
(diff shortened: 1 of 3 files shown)


pawamoy:
`_empty` is the string representation of `inspect.Signature.empty`. I think it should/could be handled somewhere else in the code. I will take a look at that later 🙂

RDIL:
We could just replace it with `None`

pawamoy:
Maybe `annotation_to_string` in `utils` should return `""` when the annotation is `inspect.Signature.Empty`. I'll try that.

pawamoy:
OK, you can undo your changes in `docstrings.py` and add these two lines in `utils.py`:

```diff
def annotation_to_string(annotation):
+    if annotation is inspect.Signature.empty:
+        return ""
```

cs01:
Sounds good. I have it marked that maintainers can edit this PR. I probably won't get to this for a while, so feel free to take this over or close it out.

PR #53, opened 2020-03-05
Do not reload numpy

jaimergp:
Temporary workaround to prevent numpy from being reloaded when live watching locally.

See issue raised at https://github.com/pawamoy/mkdocstrings/issues/36#issuecomment-594495481 for more details.


pawamoy:
Thanks :slightly_smiling_face:  !

PR #56, opened 2020-03-13
s/name/path to enable loading of nested modules

janchorowski:
Hi,

the current behavior fails to load submodules that are 3 levels deep. Say there is a module named `a.b.c`. When processing `a.b` name gets set to `b` and then `import_object(b.c)` is called, which fails. After the change `import_object(a.b.c)` is called, which works.


pawamoy:
Hi! Thanks a lot! I'll patch [py-tkdocs](https://github.com/pawamoy/py-tkdocs) as well, unless you'd like to send the patch there? Collecting documentation will be done with pytkdocs in the (close) future.

PR #69, opened 2020-04-01
Enhance some small things

RDIL:
Signed-off-by: Reece Dunham <me@rdil.rocks>

### DIFF
--- a/src/mkdocstrings/handlers/python.py
+++ b/src/mkdocstrings/handlers/python.py
@@ -11,7 +11,7 @@ from subprocess import PIPE, Popen  # nosec: what other option, more secure that
 from markdown import Markdown
 from mkdocs.utils import log
 
-from . import BaseCollector, BaseHandler, BaseRenderer, CollectionError
+from . import BaseCollector, BaseHandler, BaseRenderer, CollectionError, DataType
 
 
 class PythonRenderer(BaseRenderer):
@@ -51,7 +51,7 @@ class PythonRenderer(BaseRenderer):
     **`heading_level`** | `int` | The initial heading level to use. | `2`
     """
 
-    def render(self, data: dict, config: dict) -> str:
+    def render(self, data: DataType, config: dict) -> str:
         final_config = dict(self.DEFAULT_CONFIG)
         final_config.update(config)
 
@@ -149,7 +149,7 @@ class PythonCollector(BaseCollector):
             env=env,
         )
 
-    def collect(self, identifier: str, config: dict) -> dict:
+    def collect(self, identifier: str, config: dict) -> DataType:
         """
         Collect the documentation tree given an identifier and selection options.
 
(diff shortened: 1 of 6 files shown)


src/mkdocstrings/handlers/python.py
```diff
@@ -204,13 +204,13 @@ def collect(self, identifier: str, config: dict) -> dict:
             raise CollectionError(result["error"])
 
         if result["loading_errors"]:
-            for error in result["loading_errors"]:
-                log.warning(f"mkdocstrings.handlers.python: {error}")
+            for err in result["loading_errors"]:
+                log.warning(f"mkdocstrings.handlers.python: {err}")
```
pawamoy, line 208: I understand it helps mypy, but I really don't like that. I'd rather have a `# type: ignore` comment.

src/mkdocstrings/handlers/python.py
```diff
@@ -204,13 +204,13 @@ def collect(self, identifier: str, config: dict) -> dict:
...
 
         if result["parsing_errors"]:
-            for path, errors in result["parsing_errors"].items():
-                for error in errors:
-                    log.warning(f"mkdocstrings.handlers.python: {error}")
+            for path, problems in result["parsing_errors"].items():
+                for err in problems:
+                    log.warning(f"mkdocstrings.handlers.python: {err}")
```
pawamoy, line 213: Same remark as before: I'd rather have a `# type: ignore` comment instead of using more variables.

src/mkdocstrings/plugin.py
```diff
@@ -103,7 +107,8 @@ class MkdocstringsPlugin(BasePlugin):
 
     def __init__(self) -> None:
         super(MkdocstringsPlugin, self).__init__()
-        self.mkdocstrings_extension = None
+        # stop mypy from complaining about NoneType mismatches
+        self.mkdocstrings_extension = MkdocstringsExtension(config={})
```
pawamoy, line 111: I'd prefer a `# type: ignore` comment here as well, and reverting the change in the class init method. Having an uninitialized extension seems fragile.

src/mkdocstrings/handlers/python.py
RDIL: That was actually some other tool complaining (don't remember which) but thats fine

RDIL:
Feedback addressed 👍

src/mkdocstrings/plugin.py
```diff
@@ -103,6 +107,7 @@ class MkdocstringsPlugin(BasePlugin):
 
     def __init__(self) -> None:
         super(MkdocstringsPlugin, self).__init__()
+        # stop mypy from complaining about NoneType mismatches
```
pawamoy, line 110: Shouldn't this be removed as well :smile_cat: ?

RDIL: Whoops!  Yup I'll fix that

Issue #59, opened 2020-03-21
Make templates overridable


pawamoy:

Holt59:
I have a working version for this using a plugin configuration `custom_templates` (similar to `custom_dir` for `theme`):

```
plugins:
  - mkdocstrings:
      custom_templates: "./docs/custom_templates"
```

And then within `custom_templates`, I follow the structure you use, e.g., `python/material`.

Currently, the renderer will throw `ThemeNotSupported` even if the custom template folder contains everything needed so I am not sure how to handle that... Maybe not throw the exception? In my opinion, a warning should be sufficient since the renderer will fail later by not finding the require template.

Let me know your thoughts on this so that I can do the proper changes and then make a PR!

pawamoy:
Thank you for your interest and contribution @Holt59!

I think the best thing to do would be to pass the root of the templates folder to the jinja env's file system loader, and indeed let the renderer fail later (and catch potential exceptions to issue errors or warnings).

https://jinja.palletsprojects.com/en/2.11.x/api/#jinja2.FileSystemLoader

So instead of the whole

```python
        themes_dir = Path(__file__).parent.parent / "templates" / directory
        theme_dir = themes_dir / theme
        if not theme_dir.exists():
            if self.FALLBACK_THEME != "":
                log.warning(
                    f"mkdocstrings.handlers: No '{theme}' theme in '{directory}', "
                    f"falling back to theme '{self.FALLBACK_THEME}'"
                )
                theme_dir = themes_dir / self.FALLBACK_THEME
            else:
                raise ThemeNotSupported(theme)

        self.env = Environment(autoescape=True, loader=FileSystemLoader(str(theme_dir)))
```

...we could simply do

```python
        templates_dir = str(Path(__file__).parent.parent / "templates")
        self.env = Environment(autoescape=True, loader=FileSystemLoader(templates_dir))
```

With this change, we would need to refactor the current material templates to use

```
{% include "python/material/class.html" %}
```

instead of

```
{% include "class.html" %}
```

But this change would also allow to override templates very easily!

If the user defines a `custom_templates` key in the plugin config, we can simply pass it to the jinja loader:

```python
        templates_dir = str(Path(__file__).parent.parent / "templates")
        self.env = Environment(
            autoescape=True,
            loader=FileSystemLoader([CUSTOM_TEMPLATES_PATH, templates_dir])
        )
```

...then the custom templates will be used first if they exist :slightly_smiling_face: 

[...]

PR #82, opened 2020-04-12
WIP: Add option to allow overriding of templates.

Holt59:
From #59

### DIFF
--- a/src/mkdocstrings/extension.py
+++ b/src/mkdocstrings/extension.py
@@ -118,9 +118,9 @@ class AutoDocProcessor(BlockProcessor):
 
             handler_name = self.get_handler_name(config)
             log.debug(f"mkdocstrings.extension: Using handler '{handler_name}'")
-            handler = get_handler(handler_name,
-                                  self._config["theme_name"],
-                                  self._config["mkdocstrings"]["custom_templates"])
+            handler = get_handler(
+                handler_name, self._config["theme_name"], self._config["mkdocstrings"]["custom_templates"]
+            )
 
             selection, rendering = self.get_item_configs(handler_name, config)
 
@@ -138,8 +138,11 @@ class AutoDocProcessor(BlockProcessor):
             try:
                 rendered = handler.renderer.render(data, rendering)
             except TemplateNotFound as error:
-                log.error(f"mkdocstrings.extension: Template \"{error.name}\" not found "
-                          f"for \"{handler.renderer.directory}\" handler and theme \"{self._config['theme_name']}\".")
+                theme_name = self._config["theme_name"]
+                log.error(
+                    f"mkdocstrings.extension: Template '{error.name}' not found "
+                    f"for '{handler_name}' handler and theme '{theme_name}'."
+                )
                 return
 
             log.debug("mkdocstrings.extension: Loading HTML back into XML tree")
--- a/src/mkdocstrings/extension.py
+++ b/src/mkdocstrings/extension.py
@@ -117,7 +117,9 @@ class AutoDocProcessor(BlockProcessor):
 
             handler_name = self.get_handler_name(config)
             log.debug(f"mkdocstrings.extension: Using handler '{handler_name}'")
-            handler = get_handler(handler_name, self._config["theme_name"])
+            handler = get_handler(handler_name,
+                                  self._config["theme_name"],
+                                  self._config["mkdocstrings"]["custom_templates"])
 
             selection, rendering = self.get_item_configs(handler_name, config)
 
--- a/src/mkdocstrings/extension.py
+++ b/src/mkdocstrings/extension.py
@@ -34,6 +34,7 @@ from markdown.blockprocessors import BlockProcessor
 from markdown.extensions import Extension
 from markdown.util import AtomicString
 from mkdocs.utils import log
+from jinja2.exceptions import TemplateNotFound
 
 from .handlers import CollectionError, get_handler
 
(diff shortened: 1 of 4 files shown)


pawamoy:
Why is CI not triggering :angry:

src/mkdocstrings/handlers/__init__.py
```diff
@@ -123,7 +130,13 @@ def __init__(self, directory: str, theme: str) -> None:
...
                 raise ThemeNotSupported(theme)
 
-        self.env = Environment(autoescape=True, loader=FileSystemLoader(str(theme_dir)))
+        dirs = [str(theme_dir)]
+
+        # If the custom directory exists, we need to prepend it to the list:
+        if custom_dir is not None and custom_dir.exists():
+            dirs = [(str(custom_dir))] + dirs
+
+        self.env = Environment(autoescape=True, loader=FileSystemLoader(dirs))
```
pawamoy, line 139: Instead of all these checks, lets simply pass the three directories (custom one, theme one, fallback one) to the env loader.

Something like

```python
paths = []

if custom_templates is not None:
    paths.append(Path(custom_templates) / directory / theme)

themes_dir = Path(__file__).parent.parent / "templates" / directory

paths.append(str(themes_dir / theme))
paths.append(str(themes_dir / self.FALLBACK_THEME))

self.env = Environment(autoescape=True, loader=FileSystemLoader(paths))
```

Then we can catch the `jinja2.TemplateNotFound` exception in the extension itself.

Holt59:
I've made some changes, and I try to came up with a correct error message but let me know if it's not good for you.

src/mkdocstrings/handlers/__init__.py
```diff
@@ -110,20 +111,24 @@ def __init__(self, directory: str, theme: str) -> None:
...
-                )
-                theme_dir = themes_dir / self.FALLBACK_THEME
-            else:
-                raise ThemeNotSupported(theme)
-
-        self.env = Environment(autoescape=True, loader=FileSystemLoader(str(theme_dir)))
+
+        paths.append(themes_dir / theme)
+
+        if self.FALLBACK_THEME != "" and False:
```
pawamoy, line 128: Why the `and False`? This will never be true :confused:

src/mkdocstrings/extension.py
```diff
@@ -132,7 +135,12 @@ def run(self, parent: Element, blocks: Element) -> None:
             handler.renderer.update_env(self.md, self._config)
 
             log.debug("mkdocstrings.extension: Rendering templates")
-            rendered = handler.renderer.render(data, rendering)
+            try:
+                rendered = handler.renderer.render(data, rendering)
+            except TemplateNotFound as error:
+                log.error(f"mkdocstrings.extension: Template \"{error.name}\" not found "
+                          f"for \"{handler.renderer.directory}\" handler and theme \"{self._config['theme_name']}\".")
```
pawamoy, line 142: You can use single quotes `'` instead of escaped double-quotes `\"`.
Also you can assign `theme_name = self._config["theme_name"]` before to simplify the f-string :slightly_smiling_face:

pawamoy: The user doesn't really need to know what directory the handler is using. You could just use `handler_name` instead of `handler.renderer.directory`, and avoid assigning `directory` to `self.directory` in the base renderer's constructor.

src/mkdocstrings/handlers/__init__.py
```diff
@@ -110,20 +111,24 @@ def __init__(self, directory: str, theme: str) -> None:
         Arguments:
             directory: The name of the directory containing the themes for this renderer.
             theme: The name of theme to use.
+            custom_templates: Directory containing custom templates.
         """
+
+        self.directory = directory
```
pawamoy, line 117: As stated in a previous comment, you can drop this line and use `handler_name` instead of `handler.renderer.directory`.

pawamoy:
OK so apparently CI for PRs from forks was never a thing: https://gitlab.com/gitlab-org/gitlab/issues/5667 :crying_cat_face:

src/mkdocstrings/handlers/__init__.py
Holt59: Because I was testing something and I forgot to remove it...

src/mkdocstrings/extension.py
Holt59: Yes, I was looking for a name and I did see that `handler_name` was available. This is updated in 
e275221

src/mkdocstrings/handlers/__init__.py
Holt59: Fixed in e275221

pawamoy:
OK this is good now :slightly_smiling_face: 

Thanks a lot @Holt59 this is great :tada:!

Anything else before we merge?

Holt59:
> OK so apparently CI for PRs from forks was never a thing: https://gitlab.com/gitlab-org/gitlab/issues/5667 😿

Unfortunately `make lint` and `make check` do not work for me on Windows (issues with `black` and `scripts/*.py`). Since I only modified files under `src/`, I ran manually:

- `poetry run bandit -r src/` - Passed.
- `poetry run black --check src/` - Passed.

pawamoy:
OK no problem :slightly_smiling_face: 

Thanks a lot @Holt59 this is great :tada:!

Anything else before we merge?

Holt59:
> 
> 
> OK this is good now 🙂
> 
> Thanks a lot @Holt59 this is great 🎉!
> 
> Anything else before we merge?

No, I think it's good to go :) 

Maybe some TODOs for later (I'll open issues if necessary):

- Add a warning when the `FileSystemLoader` uses template in the fallback theme. I think this is quite complicated to do, but would be great to have.
- When there are errors (e.g., due to missing templates), the `.run()` method returns early but the `mkdocs build` process terminates correctly... Maybe there is a way to notify the build process that something wrong happened?

pawamoy:
Sure, please open issues, I'll answer on them.

pawamoy:
About the second note

> When there are errors (e.g., due to missing templates), the .run() method returns early but the mkdocs build process terminates correctly... Maybe there is a way to notify the build process that something wrong happened?

MkDocs has a `-s, --strict` flag on both `build` and `serve` command that will abort the build if there is any warning or above (errors, critical).

pawamoy:
Actually it was not hooked correctly. I'll open the corresponding issue!

Issue #91, opened 2020-04-25
Collection error v0.11.0


pawamoy:
**Describe the bug**
Collection error due to JSON failing to load the file line read on stdout. I suspect the Python interpreter to pre-execute code that is writing to stdout, or simply writing some blank lines to stdout for some reason.

**To Reproduce**
Waiting for feedback in https://github.com/rdilweb/docs/commit/cc20660bcaad8cfa114e5d197cea8cd0925a0d19

**Expected behavior**
No collection error.

**Information (please complete the following information):**
- `mkdocstrings` version: 0.11.0

**Additional context**
Possible fix:

```diff
diff --git a/src/mkdocstrings/handlers/python.py b/src/mkdocstrings/handlers/python.py
index 88e8da3..c8127eb 100644
--- a/src/mkdocstrings/handlers/python.py
+++ b/src/mkdocstrings/handlers/python.py
@@ -125,13 +125,19 @@ class PythonCollector(BaseCollector):
         """
         log.debug("mkdocstrings.handlers.python: Opening 'pytkdocs' subprocess")
         env = os.environ.copy()
+        env["PYTHONUNBUFFERED"] = "1"
 
-        python_str = "; ".join(setup_commands) + "; " if setup_commands else ""
-        cmd = python_str + "from pytkdocs.cli import main; main(['--line-by-line'])"
+        if setup_commands:
+            cmd = [
+                "python",
+                "-c",
+                "; ".join(setup_commands) + "; from pytkdocs.cli import main; main(['--line-by-line'])"
+            ]
+        else:
+            cmd = ["pytkdocs", "--line-by-line"]
 
-        env["PYTHONUNBUFFERED"] = "1"
         self.process = Popen(  # noqa: S603,S607 (we trust the input, and we don't want to use the absolute path)
-            ["python", "-c", cmd], universal_newlines=True, stderr=PIPE, stdout=PIPE, stdin=PIPE, bufsize=-1, env=env,
+            cmd, universal_newlines=True, stderr=PIPE, stdout=PIPE, stdin=PIPE, bufsize=-1, env=env,
         )
 
     def collect(self, identifier: str, config: dict) -> Any:
```

Ping @RDIL @rossmechanic

rossmechanic:
Hi @pawamoy , just tried this locally, and haven't been able to produce any errors (it has all been working perfectly, with or without setup_commands). However, I have a fix in order to get the print statements back to stdout. We can do the following to catch the stdout and flush it:
[...]

PR #103, opened 2020-05-05
Use the proper python executable to start subprocesses

RDIL:
Signed-off-by: Reece Dunham <me@rdil.rocks>

The plan is that this should fix `python` not being properly resolved, and maybe help out with #91


pawamoy:
Indeed, just tried in your docs repo on gitpod.io, this fixes the issue. The error message on gitpod.io now is "python not found", which might have been the root cause of #91 (but was hidden away somehow). Thanks!

RDIL:
That is what I was trying to fix, I recently created a custom image that doesn't have pyenv and started getting this issue because python itself isn't linked to anything, whereas python3 is

PR #112, opened 2020-05-15
Support examples section

igonro:
This PR is to support the examples section.

In progress...

Issue https://github.com/pawamoy/pytkdocs/issues/8 and PR https://github.com/pawamoy/pytkdocs/pull/35

### DIFF
--- a/src/mkdocstrings/templates/python/material/docstring.html
+++ b/src/mkdocstrings/templates/python/material/docstring.html
@@ -14,6 +14,10 @@
       {% with return = section.value %}
         {% include "return.html" with context %}
       {% endwith %}
+    {% elif section.type == "examples" %}
+      {% with examples = section.value %}
+        {% include "examples.html" with context %}
+      {% endwith %}
     {% endif %}
   {% endfor %}
 {% endif %}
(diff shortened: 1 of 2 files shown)


pawamoy:
You must use an `{% include` directive as well, like for other types of sections, as it will allow users to override only the `examples.html` template, and not the whole `docstring.html` one.

In your other PR I also explain that we should parse interactive-like Python code blocks. Here the `examples.html` template would iterate on the list of paragraphs (markdown paragraphs and code blocks), and:

- render markdown paragaph with the `convert_markdown` filter
- render code blocks with the `highlight(language="python")` filter

igonro:
This is working as expected with https://github.com/pawamoy/pytkdocs/pull/35.

I know that maybe it would be better to create a examples.html template, but I don't have knowledges about Django, neither time to get on with it.

Maybe we should add an issue with 'improvement' tag about this, but for now sadly I can't spend more time on this feature.

pawamoy:
> Django

Jinja2! :grin: 

Of course, thank you very much for the time you spent on working on this feature. I'll take it from here (though I don't have much time these days either :sweat_smile:).

Could you confirm the "allow maintainers to edit this PR" box is checked? Thanks!

igonro:
> 
> 
> > Django
> 
> Jinja2! 😁
> 
> Of course, thank you very much for the time you spent on working on this feature. I'll take it from here (though I don't have much time these days either 😅).
> 
> Could you confirm the "allow maintainers to edit this PR" box is checked? Thanks!

Yes! It is checked. Sorry for the late response.

pawamoy:
Thanks again @igonro, I'll release this soon :slightly_smiling_face:

Issue #134, opened 2020-07-13
Show full path only for direct children of a module


adunmore:
**Is your feature request related to a problem? Please describe.**

When I write a document page that includes multiple objects from a module, it would be nice to be able to include full paths only for the module's direct children.

The current renderer options aren't flexible enough for what I'm trying to do:

`show_root_full_path` results in a document that doesn't show object paths as often as I'd like. This is particularly harmful to usability when a document represents a module with many members. In this case, a single reference to the module path can be buried in a long document.

`show_object_full_path` shows too much. For example, when a class's path is already shown, it seems redundant to show the full paths of all its methods and attributes.

**Describe the solution you'd like**

A rendering option that would only include full object paths for the direct children of a module.

**Describe alternatives you've considered**

One option is to
- use `show_root_heading`
- manually identify all of a module's children

But this seems unnecessarily laborious, especially when documenting large modules.

**Additional context**
Example: 

For some module with the structure
```
- module
  - class_1
    - method_1
    - method_2
```

**Bad option 1:**
```
:::path.to.module
    rendering:
        show_root_heading: true
        show_root_full_path: true
```
Results in something like

```
path.to.module
  class_1
    method_1()
    method_2()
# too sparse
```

**Bad option 2:**
```
:::path.to.module
    rendering:
        show_object_full_path: true
        show_root_heading: true
```
Results in something like
```
path.to.module.class_1
  path.to.module.class_1.method_1()
  path.to.module.class_1.method_2()
# too verbose
```

**What I'd like:**
```
:::path.to.module
    rendering:
        option_for_perfect_full_paths: true
        show_root_headers: false
```
Resulting in
```
path.to.module.class_1
  method_1()
  method_2()
# just right :)
```

pawamoy:
Hello, thanks for the detailed feature request :slightly_smiling_face: 

So, if you have this:

```
module
  Class1
    method_a
  Class2
    method_b
```

and

```
::: path.to.module
    rendering:
      option_for_perfect_full_paths: true
      show_root_heading: false
```

You want to see

```
Module docstring.

path.to.module.Class1
  Class1 docstring.

  method_a()
    Method A docstring.

path.to.module.Class2
  Class2 docstring.

  method_b()
    Method B docstring.
```

Did I get this right?

Well this is a fair request, it shouldn't be too hard to implement. Do you want to try and send a PR?
We could call the option `show_root_members_full_path`. Then in the template you'd do something like:

```jinja2
[...]

PR #136, opened 2020-07-20
feat (python handler): Root children full path

adunmore:
Implements the Python handler option `show_root_members_full_path`:
Shows the full Python path of the direct children of the object at the
root of the documentation tree. Defaults to false.

References #134

### DIFF
--- a/src/mkdocstrings/handlers/python.py
+++ b/src/mkdocstrings/handlers/python.py
@@ -39,6 +39,7 @@ class PythonRenderer(BaseRenderer):
         "show_root_heading": False,
         "show_root_toc_entry": True,
         "show_root_full_path": True,
+        "show_root_members_full_path": False,
         "show_object_full_path": False,
         "show_category_heading": False,
         "show_if_no_docstring": False,
@@ -55,6 +56,7 @@ class PythonRenderer(BaseRenderer):
     **`show_root_toc_entry`** | `bool` | If the root heading is not shown, at least add a ToC entry for it. | `True`
     **`show_root_full_path`** | `bool` | Show the full Python path for the root object heading. | `True`
     **`show_object_full_path`** | `bool` | Show the full Python path of every object. | `False`
+    **`show_root_members_full_path`** | `bool` | Show the full Python path of objects that are children of the root object (for example, classes in a module). | `False`
     **`show_category_heading`** | `bool` | When grouped by categories, show a heading for each category. | `False`
     **`show_if_no_docstring`** | `bool` | Show the object heading even if it has no docstring or children with docstrings. | `False`
     **`show_source`** | `bool` | Show the source code of this object. | `True`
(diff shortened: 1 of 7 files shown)


pawamoy:
Alright, that's what I was expecting :slightly_smiling_face:

src/mkdocstrings/templates/python/material/attribute.html
```diff
@@ -2,10 +2,14 @@
 
   <div class="doc doc-object doc-attribute">
 
-    {% if not root or config.show_root_heading %}
+    {% if not root or config.show_root_heading%}
```
pawamoy, line 5: Nitpick: please re-add the space to reduce diff and stay consistent :stuck_out_tongue: 

Same remark in every other file!

pawamoy:
Actually this would be a "breaking" change for users using `show_object_full_path: yes`, as they would now need to also add `show_root_members_full_path: yes`, otherwise the first level of objects would not show full paths. This is not a serious breaking change, so I wonder if defaulting `show_root_members_full_path` to `None` would be worth instead.

src/mkdocstrings/templates/python/material/attribute.html
adunmore: Oops, good catch! I'll fix this.

src/mkdocstrings/handlers/python.py
```diff
@@ -55,6 +56,7 @@ class PythonRenderer(BaseRenderer):
     **`show_root_toc_entry`** | `bool` | If the root heading is not shown, at least add a ToC entry for it. | `True`
     **`show_root_full_path`** | `bool` | Show the full Python path for the root object heading. | `True`
     **`show_object_full_path`** | `bool` | Show the full Python path of every object. | `False`
+    **`show_root_members_full_path`** | `bool` | Show the full Python path of objects that are children of the root object (for example, classes in a module). | `False`
```
StephenBrown2, line 59: No longer `False` by default.

adunmore:
Hmm, something weird I've noticed - now `show_root_members_full_path` argument has three valid states:

- `None`: Root member full path behavior determined entirely by `show_object_full_path`
- `True`: Root member full path always shown, regardless of `show_object_full_path`
- `False`: Root member full path never shown, regardless of `show_object_full_path`

Is this desirable?

Another approach would be for `show_object_full_path: True` to override `show_root_members_full_path: False`. Something like 
```python
{% elif root_members %}
        {% set show_full_path = config.show_root_members_full_path or config.show_object_full_path %}
        {% set root_members = False %}
```

- Non-breaking change
- No need for a `None` default & a third state for the argument
- Seems intuitive for `show_object_full_path` to have the final say

pawamoy:
Yes, it sounds good. You convinced me :slightly_smiling_face: You can commit and push the snippet you mentioned. Thanks!

pawamoy:
Not sure what happened in CI for Windows. Let's ignore it for now.

PR #159, opened 2020-09-28
Add support for ReadTheDocs theme

pawamoy:
Issue: https://github.com/pawamoy/mkdocstrings/issues/107

### DIFF
--- a/docs/handlers/python.md
+++ b/docs/handlers/python.md
@@ -280,7 +280,8 @@ plugins:
 
 ## Recommended style (Material)
 
-Here are some CSS rules for the *Material for MkDocs* theme:
+Here are some CSS rules for the
+[*Material for MkDocs*](https://squidfunk.github.io/mkdocs-material/) theme:
 
 ```css
 /* Indentation. */
@@ -315,5 +316,57 @@ td code {
 td p {
   margin-top: 0 !important;
   margin-bottom: 0 !important;
+}
+```
+
+## Recommended style (ReadTheDocs)
+
+Here are some CSS rules for the built-in *ReadTheDocs* theme:
+
+```css
+/* Indentation. */
+div.doc-contents:not(.first) {
+  padding-left: 25px;
+  border-left: 4px solid rgba(230, 230, 230);
+  margin-bottom: 60px;
+}
+
+/* Don't use vertical space on hidden ToC entries. */
+.hidden-toc::before {
+  margin-top: 0 !important;
+  padding-top: 0 !important;
+}
+
+/* Don't show permalink of hidden ToC entries. */
+.hidden-toc a.headerlink {
+  display: none;
+}
+
+/* Avoid breaking parameters name, etc. in table cells. */
+td code {
+  word-break: normal !important;
+}
+
+/* For pieces of Markdown rendered in table cells. */
+td p {
+  margin-top: 0 !important;
+  margin-bottom: 0 !important;
+}
+
+/* Avoid breaking code headings. */
+.doc-heading code {
+  white-space: normal;
+}
+
+/* Improve rendering of parameters, returns and exceptions. */
+.field-name {
+  min-width: 100px;
+}
+.field-name, .field-body {
+  border: none !important;
+  padding: 0 !important;
+}
+.field-list {
+  margin: 0 !important;
 }
 ```
--- a/src/mkdocstrings/handlers/base.py
+++ b/src/mkdocstrings/handlers/base.py
@@ -10,6 +10,7 @@ It also provides two methods:
 """
 
 import importlib
+import re
 import textwrap
 from abc import ABC, abstractmethod
 from pathlib import Path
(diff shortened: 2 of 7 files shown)


docs/handlers/python.md
```diff
@@ -317,3 +318,55 @@ td p {
   margin-bottom: 0 !important;
 }
 ```
+
+## Recommended style (ReadTheDocs)
+
+Here are some CSS rules for the builtin *ReadTheDocs* theme:
```
RDIL, line 324:
```suggestion
Here are some CSS rules for the built-in *ReadTheDocs* theme:
```

RDIL (changes requested):

src/mkdocstrings/handlers/__init__.py
```diff
@@ -66,10 +67,44 @@ def do_highlight(
...
+
+    Returns:
+        The code properly wrapped for later highlighting by JavaScript.
+    """
+    if dedent:
+        src = textwrap.dedent(src)
+    if inline:
+        src = re.sub(r"\n\s*", "", src)
+        return f'<code class="hightlight">{do_mark_safe(src)}</code>'
+    return f'<div class="hightlight {language or ""}"><pre><code>\n{do_mark_safe(src)}\n</code></pre></div>'
```
oprypin, line 105: Typo `hightlight`?

src/mkdocstrings/handlers/__init__.py
```diff
@@ -128,10 +163,16 @@ def __init__(self, directory: str, theme: str, custom_templates: Optional[str] =
             paths.append(themes_dir / self.fallback_theme)
 
         self.env = Environment(autoescape=True, loader=FileSystemLoader(paths))  # type: ignore
-        self.env.filters["highlight"] = do_highlight
         self.env.filters["any"] = do_any
         self.env.globals["log"] = get_template_logger()
 
+        if theme == "readthedocs":
+            highlight_function = do_js_highlight
```
oprypin, line 170: readthedocs theme also seamlessly supports non-js highlight, I've been using it here
(note that _mkdocstrings_ is **not** involved in that project, just an example)

* https://oprypin.github.io/crsfml/tutorials/
* https://github.com/oprypin/crsfml/blob/5031f10351d5697e8adf8dd1bb7c781f7fdba739/tutorials/mkdocs.yml#L16

Hope that will also be supported here.

docs/handlers/python.md
```diff
@@ -317,3 +318,55 @@ td p {
   margin-bottom: 0 !important;
 }
 ```
+
+## Recommended style (ReadTheDocs)
```
oprypin, line 322: (The problem is not new in this pull request, but here goes anyway:)

Could you make this CSS available as a file, or perhaps even include it by default somehow? Otherwise it seems like users are forced to copy-paste this into their project, with the risk of it getting outdated

src/mkdocstrings/handlers/__init__.py
pawamoy: Indeed! Thanks a lot!

src/mkdocstrings/handlers/__init__.py
pawamoy: Honestly I still don't understand how the syntax highlighting is dealt with between MkDocs' built-in themes or the Material for MkDocs theme :confused: 

But yeah ideally we would support what the theme supports!

docs/handlers/python.md
pawamoy: Those are valid points. I guess I was just a bit reluctant to incorporate the CSS into the package since I suck at it :stuck_out_tongue: 

Hopefully if my CSS is bad, users will fix it in PRs, so why not! I'll see how it can be done.

src/mkdocstrings/handlers/__init__.py
```diff
@@ -66,10 +67,44 @@ def do_highlight(
     result = highlighter.highlight(src=src, language=language, linestart=line_start, inline=inline)
 
     if inline:
-        return do_mark_safe(result.text)
+        return f'<code class="highlight language-{language}">{do_mark_safe(result.text)}</code>'
```
oprypin, line 70: All of this needs to be `do_mark_safe`, otherwise the output is `&lt;code class="highlight` etc.

docs/handlers/python.md
pawamoy: Opened #189 to track this.

src/mkdocstrings/handlers/__init__.py
pawamoy: I'll users open issues about this. I don't have the time to investigate how syntax highlighting is done in all the different themes :confused: 

Let's bring initial support first and see what happens.

zillionare:
it's great to have readthedocs on, because I'd like to host my docs on readthedocs site.

When would this feature be published?

pawamoy:
Hey @zillionare, I don't have an exact date, but expect it to be release between now and January. I want to check for regressions in popular repositories using mkdocstrings before actually publishing version 0.14.0, to avoid breaking CI jobs :slightly_smiling_face:

PR #164, opened 2020-10-22
Update README.md with conda install

bandersen23:
mkdocstrings is super useful so I added it to conda-forge to make it work with my workflow easier. Figured I'd add those install instructions to the README. Thanks!

I had to add pytkdocs as well since mkdocstrings depends on it.


pawamoy:
Thanks!

PR #170, opened 2020-11-01
feat: Allow specifying heading_level as a Markdown heading instead

oprypin:
This also refactors the regex to use the `MULTILINE` flag instead of the complex "manual" alternative for detecting newlines.


pawamoy:
Hehehe, you're quick :wink: 

Sorry for not merging in order :sweat_smile:

oprypin:
There would've been a conflict regardless. I was ready :)

oprypin:
Hold on, I might have detected a buggy interaction with the passed config dict (now ChainMap but doesn't change the situation) here. This is mutating it, which may affect other things.

oprypin:
It is done.

> This actually happened to not cause any problems even with the mutation, but better to be safe.

pawamoy:
Great, thanks, squashing and merging!

PR #171, opened 2020-11-01
refactor: Use ChainMap instead of copying dicts

oprypin:


src/mkdocstrings/handlers/python.py
```diff
@@ -65,15 +66,14 @@ class PythonRenderer(BaseRenderer):
...
-        final_config.update(config)
+        final_config = ChainMap(config, self.default_config)
 
         template = self.env.get_template(f"{data['category']}.html")
 
         # Heading level is a "state" variable, that will change at each step
         # of the rendering recursion. Therefore, it's easier to use it as a plain value
         # than as an item in a dictionary.
-        heading_level = final_config.pop("heading_level")
+        heading_level = final_config["heading_level"]
```
pawamoy, line 76: I guess we can't pop from a `ChainMap`? Anyway, I also think it doesn't hurt to let `heading_level` in the final config, as the original value :+1:

oprypin: You can technically `pop`, but really it gets forwarded to the first dict, which is a totally different operation. E.g. if both dicts have the key, then it'd just make the value of the 2nd one visible instead...

pawamoy: Okay, thanks for the explanation, LGTM.

PR #172, opened 2020-11-01
fix: Detecting paths relative to template directory in logging

### LINKED ISSUE #166: Not Windows-compatible code in loggers
**Describe the bug**

In https://github.com/pawamoy/mkdocstrings/blob/master/src/mkdocstrings/loggers.py#L107:

```python
return template.filename.rsplit("/templates/", 1)[1]
```

The forward slashes are not cross-platform :wink:

**Additional context**
We should probably use `os.sep` instead of `/` here.

oprypin:
Fixes #166

### DIFF
--- a/src/mkdocstrings/loggers.py
+++ b/src/mkdocstrings/loggers.py
@@ -1,11 +1,14 @@
 """Logging functions."""
 
 import logging
+from pathlib import Path
 from typing import Callable, Optional
 
 from jinja2 import contextfunction
 from jinja2.runtime import Context
 from mkdocs.utils import warning_filter
+
+from mkdocstrings.handlers import base
 
 
 class LoggerAdapter(logging.LoggerAdapter):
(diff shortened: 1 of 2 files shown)


src/mkdocstrings/loggers.py
```diff
@@ -102,9 +105,12 @@ def get_template_path(context: Context) -> str:
     Returns:
         The relative path to the template.
     """
-    template = context.environment.get_template(context.name)
-    if template.filename:
-        return template.filename.rsplit("/templates/", 1)[1]
+    filename = context.environment.get_template(context.name).filename
+    if filename:
+        try:
+            return str(Path(filename).relative_to(base.TEMPLATES_DIR))
```
pawamoy, line 111: I'm not sure this would work for custom templates, which would exist outside of `base.TEMPLATES_DIR`?

oprypin: Yes, so an absolute path will be used for those. It's good because otherwise it'd be ambiguous.

pawamoy: Oh, perfect then :smile:

pawamoy:
Thanks again! :tada:

PR #185, opened 2020-12-02
fix: xreferences from the root index page

oprypin:
If there's any xref inside docs/index.md, `url_a` ends up being '', and `url_a[-1]` fails.


pawamoy (approved):
Nice refactor! Thanks a lot!

PR #188, opened 2020-12-06
Rework autorefs replacement to not re-parse the final HTML

### LINKED ISSUE #187: `fix_refs` is by far the slowest part of this
It's quite sad that I'm running into real slowness when trying to apply mkdocstrings.

Anyway, I ran some profiling on a site with a lot of doc objects, and found that over half of the time (of the whole site build) is spent in `fix_refs`. It's probably not a surprise that it's slow, but it's really *that* slow. So if anyone's looking to improve performance, look at this first.

image

oprypin:
Define a "transport representation" for autorefs:

```html
<span data-mkdocstrings-identifier="SomeIdentifier">SomeHTML</span>
```

First a native Markdown extension translates from the usual `[SomeMarkdown][SomeIdentifier]` to the above, and then the post-process replacement mechanism (which is kept in the same place as before) doesn't need to be so careful and complicated, it just indiscriminately replaces such exact strings.

This is a very big boost in performance and I think is more future-proof.

Other mkdocstrings handlers are also free to use this mechanism: define whatever syntax for autorefs that they need and then insert this exact HTML themselves, for the postprocessing to pick up later. It used to be possible to insert the square-brackets Markdown before, but that was very fragile.

Fixes #187.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -11,7 +11,6 @@ These projects were used to build `mkdocstrings`. **Thank you!**
 
 ### Direct dependencies
 [`autoflake`](https://github.com/myint/autoflake) |
-[`beautifulsoup4`](http://www.crummy.com/software/BeautifulSoup/bs4/) |
 [`black`](https://github.com/psf/black) |
 [`coverage`](https://github.com/nedbat/coveragepy) |
 [`failprint`](https://github.com/pawamoy/failprint) |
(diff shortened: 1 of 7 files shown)


oprypin:
Fixes #187.

Compare the performance profile to what I posted there.

image

pawamoy (approved):
Well @oprypin, this is absolutely fantastic! Thank you so much!

Less code, more robust, future-proof, new tests, and you got rid of beautifulsoup... this is great :rocket: !

Issue #179, opened 2020-11-13
Is there any way to skip creating the markdown files and nav manually ?


aahnik:
mkdocstrings supports the injection as said in docs, 
```
::: my_library.my_module.my_class
```

I also have to create the navigation.

This is great because it gives precise control, but is there any existing way to completely generate all the pages with sane defaults?

Like each module's docs in separate page. And links to sub or super module. Something like [pdoc3](https://pdoc3.github.io/pdoc/) but using the material theme.

Many thanks for your help

pawamoy:
I've been using this ugly Bash script myself:

```bash
#!/usr/bin/env bash
cd src || exit 1
rm -rf ../docs/reference
mkdir ../docs/reference
find your_package | grep -v -e pyc\$ -e __pycache__ -e '^your_package$' -e py.typed -e __init__ -e __main__ | sort | while read -r line; do
    indent="$(echo "${line}" | sed -r 's/\w+\//  /g;s/[^ ]+$//')"
    case "${line}" in
        *.py)
            mod=${line##*/}
            line=${line%.py}
            dot=${line//\//.}
            md=${line#r_pytb/platform/}.md
            dir=${md%/*}
            if [ ! "${dir}" = "${md}" ]; then
                mkdir -p ../docs/reference/${dir}
            fi
            echo "::: ${dot}" > ../docs/reference/${md}
            echo "${indent}- ${mod}: reference/${md}"
        ;;
        *)
            echo "${indent}- ${line##*/}:"
    esac
done
```

It will regenerate the `reference/` directory in the `docs/` one, and also output the related `nav` for `mkdocs.yml`.

We could consider allowing handlers to provide ways to do this (creating files with `:::` and outputing/updating the nav) and provide a common CLI for all handlers.

aahnik:
@pawamoy Thanks a lot.

I was planning to write somekind of script. The script you shared, will perfectly do my work for now.

Having some option in the cli would be great ! ☺️

oprypin:
This will be even more relevant for [Crystal handler](https://github.com/oprypin/mkdocstrings-crystal) because Crystal programming language tends to have a ton of classes, and the convention for docs currently is to put one class per doc page.

See my script for generation in the commit message here:
https://github.com/oprypin/athena-website/commit/264e549303477a7c374a8c0ca43a7a6742efab08 😩

[...]

PR #191, opened 2020-12-07
refactor: Split out the handler cache, expose it through the plugin

oprypin:
This makes `handlers_cache` no longer be global but instead be confined to the Plugin. There will be only one instance of the plugin so it doesn't matter anyway. But actually this is also more correct, because what if someone tried to instantiate multiple handlers with different configs? It would work incorrectly previously.

But my main goal for this is to expose `MkdocstringsPlugin.get_handler(name)`. Then someone can use this inside a mkdocs hook:

```python
def on_files(self, files: Files, config: Config):
    docs = config['plugins']['mkdocstrings'].get_handler('python').collector
```

So this is basically a prerequisite for issue #179: one could query the collector to know which files to generate.

### DIFF
--- a/src/mkdocstrings/extension.py
+++ b/src/mkdocstrings/extension.py
@@ -35,7 +35,7 @@ from markdown.blockprocessors import BlockProcessor
 from markdown.extensions import Extension
 from markdown.util import AtomicString
 
-from mkdocstrings.handlers.base import CollectionError, get_handler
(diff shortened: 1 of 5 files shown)


pawamoy (approved):
Looking good :rocket:! Can't wait to see what you do next :drooling_face: Thanks a lot!

Issue #193, opened 2020-12-10
Content tabs from different modules in the same markdown cross link to the first one always


zuru:
Not sure if this is related to mkdocs, mkdocstrings, or pytkdocs.
When using content tabs with the Tabbed extension:

``` yaml
markdown_extensions:
  - pymdownx.tabbed
```

within different module docstrings, which are rendered on the same page, their IDs are regenerated and the following content tabs link to the first content tabs when clicked.

To reproduce, add a content tab block in the docstring of two modules:

``` yaml
=== "Tab1"
   some content

=== "Tab2"
   some content2

```

and then add the modules in the same markdown:

``` yaml
::: module1

::: module2

```

pawamoy:
Hello @zuru, thank you for the report!

I was able to confirm this is a bug, unfortunately I'm going to mark this as "won't fix" for now because we (@oprypin and I) don't see an obvious solution to this.

To summarize, the bug happens because we render each Markdown docstring separately, so we can't keep track of previous tabs with same title and therefore they end up with duplicated IDs. The last tabs then always redirect to the first ones with same IDs.

@oprypin has been fixing similar bugs and implementing similar features using custom Markdown extensions, but we fear (correct me if I'm wrong @oprypin!) this issue would need weird hacks that would complicate the code too much. I don't really want to alter the whole Markdown rendering process, who knows what other issues stems from the fact we render each docstring separately...

So for now, what I did was add a note in the troubleshooting section of the docs (still not deployed) asking users to avoid using identical titles for different tabs that will be rendered in the *same* markdown page. Emphasis on *same* Markdown page because this issue does not happen when rendering tabs with identical titles on *different* pages.

I understand this can be annoying, but we don't have a solution for this yet :confused: 
Don't hesitate to share more on this subject of course!

zuru:
Hi @pawamoy, thanks for checking it out. As a quick workaround, we used this js script to fix it:

```js
///tab incremental id// 
function setID(){
    var tabs = document.getElementsByClassName("tabbed-set");
    for (var i = 0; i < tabs.length; i++) {
        children = tabs[i].children;
        var counter = 0;
        var iscontent = 0;
        for(var j = 0; j < children.length;j++){
            if(typeof children[j].htmlFor === 'undefined'){
                if((iscontent + 1) % 2 == 0){
                    //check if it is content
                    if(iscontent == 1){
                        btn = children[j].childNodes[1].getElementsByTagName("button");
                    }
                }
                else{
[...]

PR #195, opened 2020-12-12
Fix lack of isolation of Markdown calls from the top-level Markdown

### LINKED ISSUE #186: Footnote appears in all following objects after the first appearance
Hi, I'm not sure this is related to pytkdocs or mkdocstrings.

**Describe the bug**
If a footnote is added to one of python objects, the footnote will appear in every object after that one in the same documentation page.  

**To Reproduce**
Here is an example to reproduce

```python
# example.py
def C():
    """This is C
    """
    pass

def B():
    """This is B[^1]

    [^1]: This is footnote
    """
    pass

def A():
    """This is A
    """
    pass
````
```
# mkdocs.yml
site_name: Footnote
theme:
  name: "material"
plugins:
  - search
  - mkdocstrings:
      default_handler: python
markdown_extensions:
  - footnotes
```
**Expected behavior**
The footnote is supposed to appear only under function B

**Screenshots**
However, footnote appears under function B and C

[
Screenshot from 2020-12-05 00-30-54
](url)

### LINKED ISSUE #192: Markdown headings can bust to the topmost nav level
I'm actually not entirely sure if it's the case for Python renderer, but in my own renderer I ran into the issue that if the docstring contains a `# Markdown Heading`, it's not adjusted according to the current `heading_level`, so in the nav it'll just appear as a topmost item regardless of what's happening around it.

I'm currently looking into a solution.

oprypin:
Fixes #186
Fixes #192


pawamoy:
Thanks a lot @oprypin! You're doing great work here :star_struck: 

So, footnotes will now appear only once, but if I understand correctly, one cannot use the same id for different footnotes in different docstrings if one's plan is to render these docstrings in the same page? If yes, I'm OK with that, and we should simply document it!

pawamoy:
Also sorry for force-pushing, this is not good practice.

pawamoy:
One more thing: I don't think it fixes #193, right? For footnotes we can force different IDs, so I guess for tabs one would need to use different titles if they plan on rendering them in the same page?

oprypin:
You are correct in all regards in your comments. Yes, I'm not handling ID collisions.

And actually yes, I'm not sure at all why you force-pushed :D.
Didn't need to rebase here at all.

Do you plan to merge this as a branch, or squash perhaps?

oprypin:
ID collisions are possible to handle by prefixing them, just don't know much hackery that would require.

pawamoy:
I rebased to reword the commits :expressionless: I can be obsessive over little details.
I also didn't want to squash because this PR does several things and so I wanted to keep each commit :slightly_smiling_face: 

> ID collisions are possible to handle by prefixing them, just don't know much hackery that would require.

OK, then let's simply document it for now. I'll add a troubleshooting item, merge, and link to the docs in the related issues. Does it sound good?

oprypin:
> I rebased to reword the commits

Ah I see, didn't catch that. Thanks for the style fix as well.

Yes, all good then, no outstanding comments from me.

PR #196, opened 2020-12-12
Don't limit CI runs to just master branch

oprypin:
GitHub actions are limited only to the branch that they are defined on anyway. So all this did was to prevent CI from running on a dev/feature branch


pawamoy:
Thanks

PR #197, opened 2020-12-19
fix: More lenient regex for data-mkdocstrings-identifier

oprypin:
First introduced in https://github.com/pawamoy/mkdocstrings/pull/188, there's a regression: if something messes with the final HTML before mkdocstrings plugin gets to it, it can't be detected and replaced.
A known case is with 'minify' plugin if it appears before our plugin in the config (which it shouldn't anyway).
  
So workaround this specific case. Make the attribute quotes optional to also catch minified HTML.
Hard to do much else, though perhaps a warning would make sense.


pawamoy:
Thanks a lot! :rocket:

Issue #176, opened 2020-11-03
Support for other python docstring styles


jnoortheen:
**Is your feature request related to a problem? Please describe.**
It will be great if other standard docstring formats like `Numpy` and `RST` supported out of the box. 

**Describe the solution you'd like**
This is fairly trivial. There are libs out there like https://github.com/rr-/docstring_parser to help parse functions.

pawamoy:
Feature request for Numpy format: https://github.com/pawamoy/pytkdocs/issues/7
Pull request for RST format: https://github.com/pawamoy/pytkdocs/pull/71

RST will never be fully supported, see the discussion in the mentioned above.

Also, I had https://github.com/rr-/docstring_parser in my radar, unfortunately we do a bit more in the pytkdocs parsers: we use type annotations to populate the types of the documented attributes/arguments in docstrings, which the linked project does not do. But if you want to open a PR on pytkdocs for the Numpy format using this lib, I could review it :slightly_smiling_face: 

Leaving this issue open for visibility!

jnoortheen:
Oh I haven't checked that repo. I don't have much time lately. I converted my code to use google docstrings :) Since there are people working on it, feel free to close this issue.

pawamoy:
Since the RST PR was merged and there's an issue for the Numpy format in pytkdocs, I'll close this one.

PR #198, opened 2020-12-20
feat: Support for reStrcuturedText style

plannigan:
Updates version of `pytkdocs` to allow use of release that supports reStructuredText. Also update documentation.

Addresses #176


pawamoy:
Hello! Thank you for the PR.

This is not technically a feature so I'll squash and use a `docs` type for the commit.
Could you also revert the pytkdocs version range upgrade in pyproject.toml? I'd like to do that in a separate commit once I'm ready to release mkdocstrings version 0.14.

Thanks again!

plannigan:
I can do that. But just to be clear, I updated the docs to include a reStructuredText snippet. So the documentation will not build successfully until the new version of `pytkdocs` is pulled into the project.

The CI failure seems to be an issue `poetry` decides to use `pytkdocs==0.9.0` instead of the newer version. I'm still trying to figure out why that is happening. It might be a conflict with the allowed ranges for the indirect dependencies.

pawamoy:
> But just to be clear, I updated the docs to include a reStructuredText snippet. So the documentation will not build successfully until the new version of pytkdocs is pulled into the project.

OK sure, not a problem :slightly_smiling_face: 

> The CI failure seems to be an issue poetry decides to use pytkdocs==0.9.0 instead of the newer version. I'm still trying to figure out why that is happening. It might be a conflict with the allowed ranges for the indirect dependencies.

Hmmmm yeah, I wonder how deps would resolve if one of mkdocstrings dev dep is also depending on mkdocstrings :sweat_smile:

plannigan:
Made the change as requested. Also, I believe the linked PR will resolve the issue that showed up in CI for this PR.

Because of the inter-dependencies here, it is OK if this PR hangs out open until you are closer to releasing the new version of this project.

PR #199, opened 2020-12-22
Make ids of Markdown sub-documents prefixed with the parent item id

### LINKED ISSUE #186: Footnote appears in all following objects after the first appearance

### LINKED ISSUE #193: Content tabs from different modules in the same markdown cross link to the first one always

oprypin:
Fixes #193
https://github.com/pawamoy/mkdocstrings/issues/193#issuecomment-749134463

Fixes #186 
https://github.com/pawamoy/mkdocstrings/issues/186#issuecomment-743469671

### DIFF
--- a/src/mkdocstrings/handlers/base.py
+++ b/src/mkdocstrings/handlers/base.py
@@ -380,7 +380,7 @@ class _IdPrependingTreeprocessor(Treeprocessor):
         super().__init__(md)
         self.id_prefix = id_prefix
 
-    def run(self, root: Element):
+    def run(self, root: Element):  # noqa: WPS231 (not complex)
         if not self.id_prefix:
             return
         for el in root.iter():
--- a/src/mkdocstrings/handlers/base.py
+++ b/src/mkdocstrings/handlers/base.py
@@ -128,7 +128,7 @@ def do_any(seq: Sequence, attribute: str = None) -> bool:
     return any(_[attribute] for _ in seq)
 
 
-def do_convert_markdown(md: Markdown, text: str, heading_level: int) -> Markup:
+def do_convert_markdown(md: Markdown, text: str, heading_level: int, html_id: str = "") -> Markup:
     """
     Render Markdown text; for use inside templates.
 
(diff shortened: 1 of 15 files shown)


oprypin:
Check out this test site with both tabs and footnotes with same ids working correctly: 
https://gist.github.com/6d82074d3539851e6eaec6dc8db72dd2

This also fixes the issue I had mentioned where same headings in different docstrings would share the same id.

`#a.foo`, `#examples`, `#b.foo`, `#examples` becomes
`#a.foo`, `#a.foo--examples`, `#b.foo`, `#b.foo--examples`

oprypin:
View test site: http://pryp.in/p/6d82074d3539851e6eaec6dc8db72dd2

src/mkdocstrings/handlers/base.py
```diff
@@ -369,6 +375,53 @@ def teardown(self):
         self._handlers.clear()
 
 
+class _IdPrependingTreeprocessor(Treeprocessor):
+    def __init__(self, md, id_prefix: str):
+        super().__init__(md)
+        self.id_prefix = id_prefix
+
+    def run(self, root: Element):
```
pawamoy, line 383:
```suggestion
    def run(self, root: Element):  # noqa: WPS231 (not complex)
```

pawamoy:
Once again, fantastic contribution @oprypin :rocket: Thanks!
LGTM, lets see how it behaves in the regressions workflow :slightly_smiling_face:

oprypin:
I notice that with this change, it is now possible link to stuff like `[foo][some.identifier--subdoc-heading]`, and people are figuring it out. Should we embrace this or lock it down?

pawamoy:
I think we should embrace it, this is a pretty nice feature. What's your opinion?

oprypin:
Overall I support it, I'm just disgruntled by the fact that I didn't realize this effect when I developed this PR, and also the choice of the "`--`" separator wasn't with this usage in mind, although maybe there isn't anything better.

oprypin:
For example, maybe people would've wanted to use `some.identifier#subdoc-heading` - but on the other hand, I'm already using the `#` symbol for identifiers in Crystal language, so it'd be ambiguous.

pawamoy:
I like `--`. Headings IDs get slugified with `-` as separators, so a double `--` makes sense to me to delimit the object identifier and the subheading. A bit like Django's ORM parameters using double underscores to avoid mistaking for another attribute: `attribute__le`.

PR #200, opened 2020-12-22
Disable Jinja's auto_reload, to reduce disk reads

oprypin:
Any time a template is referenced, even though it's cached in memory, Jinja ends up doing a filesystem access, to check whether the template file has been modified and so would need to be recompiled.
But it's not even useful to be able to modify a template file on disk in the middle of a mkdocs build.

[An equivalent change to MkDocs itself proved to improve performance significantly](https://github.com/mkdocs/mkdocs/pull/2261). I can't say the same about mkdocstrings, as its templates aren't as nested, but it'll still help. And just good for consistency now.


pawamoy (approved):
:+1:

PR #201, opened 2020-12-26
perf: Call `update_env` only once per `Markdown` instance

oprypin:
The follow-up calls for each block make no difference at all, and are getting quite expensive.


pawamoy:
Easy change for improved perfs, merging! Thanks @oprypin!

PR #204, opened 2020-12-30
fix: Remove strange 'href' from <h1..6>

oprypin:
(there's no real issue with this but it's invalid and doesn't do anything)


pawamoy:
Indeed, not sure when and why I added those. Thank you!

PR #206, opened 2020-12-30
refactor: remove some unused code

oprypin:
I conclude that it's unused because the condition necessarily requires finding something with class 'autodoc' -- but there is nothing that creates anything with class 'autodoc'!


pawamoy:
Seems like you're right! It comes from the code I initially copied from mkautodoc :sweat_smile:  Thanks!

PR #207, opened 2020-12-30
fix: Don't wrongly discard text immediately before an autodoc

oprypin:
```markdown
foo
**foo**
::: some.identifier
```

In the previous state, the "foo" text would be completely obliterated, but now it will be kept, and the autodoc will be rendered as well.

### DIFF
--- a/src/mkdocstrings/extension.py
+++ b/src/mkdocstrings/extension.py
@@ -148,6 +148,8 @@ class AutoDocProcessor(BlockProcessor):
         match = self.regex.search(str(block))
 
         if match:
+            if match.start() > 0:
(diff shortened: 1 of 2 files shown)


oprypin:
[touché](https://github.com/Python-Markdown/markdown/pull/1092)

pawamoy:
This is great!

I guess it's an in-house workaround for Markdown versions lower than the next one containing your fix.

oprypin:
No no, this is just another issue, one that is exactly parallel to the linked one, just that we happened to notice the linked one first

pawamoy:
Oh OK, thank you for the explanation :slightly_smiling_face:

PR #208, opened 2020-12-31
tests: Run teardown for Handlers, refactor into contextmanager

oprypin:
This fixes random test failures that for some reason showed up only on Windows Python 3.6


pawamoy:
Perfect, thanks a lot @oprypin!

PR #209, opened 2020-12-31
refactor: Don't re-parse the whole subdoc, expose only headings

oprypin:
* ## refactor: Merge md extensions into one, to reduce boilerplate

* ## refactor: Pass the actual used Markdown instance to update_env
    
    The fact that we're re-instantiating Markdown used to be seen as a hack, but now it's required and preferred.

* ## refactor: Move do_convert_markdown method into the class

* ## refactor: Don't re-parse the whole subdoc, expose only headings
    
    Previously the autodoc sub-document (HTML with nested Markdown) was being re-integrated into the top-level document by the way of parsing it as XML and inserting it back into the XML tree of the top-level markdown duri>
    This gave a lot of side effects that need to be worked around. XML parsing, Markdown double-processing (with only situational workaround possible), etc.
    
    Now, the document will be inserted as fully opaque HTML, so those issues are avoided entirely. But now there are new issues to overcome!
    
    * The ToC extension can no longer see through this subdocument, so no ToC entries can be generated. I thought a lot how to re-add these properly, and there's no easy solution without parsing the HTML. In the end I deci>
    
    * MkDocs can no longer fix relative paths in this subdocument. That one is much easier: just manually re-add its treeprocessor.

* ## refactor: Actually exclude hidden headings from the doc
    
    Now we separately control which headings appear in the doc and which in the ToC. No need to hide them with CSS now.

### DIFF
--- a/docs/css/mkdocstrings.css
+++ b/docs/css/mkdocstrings.css
@@ -10,17 +10,6 @@ h5.doc-heading {
   text-transform: none !important;
 }
 
-/* Don't use vertical space on hidden ToC entries. */
-.hidden-toc::before {
-  margin-top: 0 !important;
-  padding-top: 0 !important;
-}
-
-/* Don't show permalink of hidden ToC entries. */
-.hidden-toc a.headerlink {
-  display: none;
-}
-
 /* Avoid breaking parameters name, etc. in table cells. */
 td code {
   word-break: normal !important;
--- a/docs/handlers/python.md
+++ b/docs/handlers/python.md
@@ -335,17 +335,6 @@ h5.doc-heading {
   text-transform: none !important;
 }
 
-/* Don't use vertical space on hidden ToC entries. */
-.hidden-toc::before {
-  margin-top: 0 !important;
-  padding-top: 0 !important;
-}
-
-/* Don't show permalink of hidden ToC entries. */
-.hidden-toc a.headerlink {
-  display: none;
-}
-
 /* Avoid breaking parameters name, etc. in table cells. */
 td code {
   word-break: normal !important;
@@ -368,17 +357,6 @@ div.doc-contents:not(.first) {
   padding-left: 25px;
   border-left: 4px solid rgba(230, 230, 230);
   margin-bottom: 60px;
-}
-
-/* Don't use vertical space on hidden ToC entries. */
-.hidden-toc::before {
-  margin-top: 0 !important;
-  padding-top: 0 !important;
-}
-
-/* Don't show permalink of hidden ToC entries. */
-.hidden-toc a.headerlink {
-  display: none;
 }
 
 /* Avoid breaking parameters name, etc. in table cells. */
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,6 +93,7 @@ exclude = ["tests/fixtures"]
     "-F821",  # redundant with E0602 (undefined variable)
     "-Q000",  # black already deals with quoting
     "-S101",  # use of assert
+    "-S405",  # we are not parsing XML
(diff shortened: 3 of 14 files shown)


oprypin:
The lint failures are interesting.
https://github.com/oprypin/mkdocstrings/runs/1639842955
I'm using the leading underscore to designate things that are private to our whole code package, but the linter insists that this makes the things private to the class.
Maybe it's a design problem but I fail to see a simple solution. Maybe you'll find one.

pawamoy:
Yes usually an underscore in front of an attribute means it should be accessed from within its instance only (with `self._attribute`), and never from somewhere else (`self.intermediate_instance._attribute`). I think it's just annoying when you design your library to be mostly private, and expose only a few things publicly. If you would like to start doing that with mkdocstrings (as you noticed everything was public in my code), then we can decide to ignore these warnings.

oprypin:
[**Regressions run**](https://github.com/oprypin/mkdocstrings-regressions/runs/1640121711) is interesting.

Something weird with HTML escaping, will need to check that out.

[Note that I'm filtering out the intentional addition of `div.mkdocstrings` and removal of `.hidden-toc`](https://github.com/oprypin/mkdocstrings-regressions/commit/xhead)
\- but see them in [a non-filtered run here](https://github.com/oprypin/mkdocstrings-regressions/runs/1639857965)

pawamoy:
Could be [`autoescape=True`](https://github.com/pawamoy/mkdocstrings/pull/209/files#diff-288aa0bb4705e3681add5ce91b76e798bcf7b7ba2570563ec6e77809fa2bc166R171): converting the templates to HTML with Jinja escapes a first time, and then during Markdown conversion it is escaped a second time?

oprypin:
This is ready now.
[Regressions run](https://github.com/oprypin/mkdocstrings-regressions/actions/runs/484312378)

oprypin:
[Latest regression run](https://github.com/oprypin/mkdocstrings-regressions/actions/runs/490833775) still looking very good.
No change from the previous one, but I'm no longer filtering out `div.mkdocstrings` because I got rid of the addition.


There's a diff that I hadn't realized before what it means
https://github.com/oprypin/mkdocstrings-regressions/runs/1715267244?check_suite_focus=true#step:6:1086

This is actually an accidental bug fix! I think previously Markdown was doing something weird to this content on the external pass, messing with newlines.
You can see it broken here 
https://servir-mekong.github.io/hydra-floods/thresholding/#hydrafloods.thresholding.kmeans_extent

but now it's fixed because we "protect" the inner HTML from the outer Markdown processing.

pawamoy:
Hey! Indeed, last time I was wondering about all these `<br />` disappearing in diffs. Well, if it's better now, great! Should I wait a bit more or do you think this is good to be merged?

pawamoy:
Also, one question: is there a reason we have `update_env` and `_update_env`? Couldn't we merge both into one?

oprypin:
You can look at the individual commit that splits update_env into two parts. It is very intentional. One part is the one that must not be overridden, and the other part becomes very friendly to overriding, because it receives a pre-constructed inner Markdown instance (unlike previously where it received the useless outer Markdown instance)

oprypin:
It is ready to be merged indeed (other than the merge conflict which I don't have my computer at the moment to resolve).
Merging is strongly recommended without squashing btw.

oprypin:
Merged. BTW another thing about update_env. I think maybe the outer `_update_env` should just get a new name, then it'd make more sense.I jusy didn't think of that.

pawamoy:
Great refactor! Thanks a lot :slightly_smiling_face:

PR #210, opened 2021-01-01
style: Remove wemake-python-styleguide, upgrade isort

oprypin:


oprypin:
Removed dependencies:

- astor
- docutils
- eradicate
- flake8-broken-line
  - who cares
- flake8-commas  
  - black handles it better
- flake8-debugger
- flake8-eradicate
  - dubious
- flake8-isort  
  - check formatting is done separately
- flake8-quotes  
  - black handles it better
- flake8-rst-docstrings  
  - no rst
- restructuredtext-lint
- testfixtures
- wemake-python-styleguide

duties.py
```diff
@@ -355,7 +355,7 @@ def format(ctx):  # noqa: W0622 (we don't mind shadowing the format builtin)
         title="Removing unused imports",
         pty=PTY,
     )
-    ctx.run(f"isort -y -rc {PY_SRC}", title="Ordering imports", pty=PTY)
+    ctx.run(f"isort {PY_SRC}", title="Ordering imports", pty=PTY)
```
pawamoy, line 358: I guess version 5 now reorders imports in python files by default, without the need for these options?

oprypin: Yes https://pycqa.github.io/isort/docs/upgrade_guides/5.0.0/#migrating-cli-options

pawamoy:
Hmmmm, I'm half-convinced about this, but I guess this is a reasonable thing to do (removing WPS and all these useless dependencies). I still feel like many WPS' warnings were useful (even if the rest were annoying): I learned a few things while using it. I would also hate to nitpick myself for future PRs instead of letting the quality step do the job. Maybe the smart thing would be to export the interesting rules of WPS into a lightweight package without all these dependencies!

Let see how it goes without. Just a question about isort.And thanks for the PR :slightly_smiling_face:

duties.py
pawamoy: Yup, got my answer in the changelog: https://github.com/pycqa/isort/blob/master/CHANGELOG.md#500-penny---july-4-2020

PR #212, opened 2021-01-03
fix: Don't mutate the original Markdown config for permalinks

oprypin:
This was a regression from https://github.com/pawamoy/mkdocstrings/pull/203.
This ended up disabling permalinks outright for all docs.

### DIFF
--- a/src/mkdocstrings/handlers/base.py
+++ b/src/mkdocstrings/handlers/base.py
@@ -232,7 +232,10 @@ class BaseRenderer(ABC):
         # Prevent a bug that happens due to treeprocessors running on the same fragment both as the inner doc and as
         # part of the re-integrated doc. Namely, the permalink '¶' would be appended twice. This is the only known
         # non-idempotent effect of an extension, so specifically prevent it on the inner doc.
-        configs.setdefault("toc", {})["permalink"] = False
+        try:
+            configs["toc"] = dict(configs["toc"], permalink=False)
+        except KeyError:
+            pass
 
         md = Markdown(extensions=extensions, extension_configs=configs)
 
--- a/tests/test_extension.py
+++ b/tests/test_extension.py
@@ -1,7 +1,9 @@
 """Tests for the extension module."""
+import copy
 from contextlib import contextmanager
 from textwrap import dedent
 
+import pytest
 from markdown import Markdown
 
 from mkdocstrings.extension import MkdocstringsExtension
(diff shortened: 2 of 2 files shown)


oprypin:
[Regressions run against 0.13.6 without this change](https://github.com/oprypin/mkdocstrings-regressions/runs/1639999217)
[Regressions run against 0.13.6 **with this change**](https://github.com/oprypin/mkdocstrings-regressions/runs/1640249641)

src/mkdocstrings/handlers/base.py
```diff
@@ -232,7 +232,10 @@ def update_env(self, md: Markdown, config: dict) -> None:
         # Prevent a bug that happens due to treeprocessors running on the same fragment both as the inner doc and as
         # part of the re-integrated doc. Namely, the permalink '¶' would be appended twice. This is the only known
         # non-idempotent effect of an extension, so specifically prevent it on the inner doc.
-        configs.setdefault("toc", {})["permalink"] = False
+        try:
+            configs["toc"] = dict(configs["toc"], permalink=False)
+        except KeyError:
+            pass
```
pawamoy, line 238: Shouldn't we simply deep copy `configs` at line 231 (before the comment)? If we find other non-idempotent extensions, we'll have to write a try except for each, no? I find the deep copy easier to reason about and less prone to future errors, what do you think?

oprypin: deepcopy is too big of a hammer. You never know what things will end up being there. I think this is clearly better for a stable release. And long-term, this will not be a thing. I'm dropping all of this code here https://github.com/pawamoy/mkdocstrings/pull/209/files#diff-288aa0bb4705e3681add5ce91b76e798bcf7b7ba2570563ec6e77809fa2bc166L232  (and the reason that I noticed this regression is that that refactor avoids this regression :smile: )

pawamoy:
> Regressions run against 0.13.6 with this change

Hmmm, `code` tag in a `code` tag? How does it compare with 0.14.0b1?

src/mkdocstrings/handlers/base.py
pawamoy: Oh, OK, makes sense, thanks

oprypin:
0.14.0b1 also has that https://github.com/oprypin/mkdocstrings-regressions/runs/1639999217#step:6:908

oprypin:
`(cd ~/repos/mkdocstrings && git bisect start HEAD 0.13.6 && git bisect run bash -c 'cd ~/repos/mkdocstrings-regressions/repos/copier-org-copier && PYTHONPATH=~/repos/mkdocstrings/src venv/bin/mkdocs build || exit 125; ! grep "data: <code class=\"highlight language-python\">" site/reference/config/objects/index.html')`

1028115682ed0806d6570c749af0e382c67d6120 is the first bad commit

oprypin:
The PR adds `<code` here
https://github.com/pawamoy/mkdocstrings/commit/1028115682ed0806d6570c749af0e382c67d6120#diff-288aa0bb4705e3681add5ce91b76e798bcf7b7ba2570563ec6e77809fa2bc166R71

and removes it from two other templates
https://github.com/pawamoy/mkdocstrings/commit/1028115682ed0806d6570c749af0e382c67d6120#diff-def19ef968f8d513fc3bd0a749a36d27d2d6868a666d96de68d2608774b4e521

but not from attribute
https://github.com/pawamoy/mkdocstrings/blob/3ee6bf16839e4a79e04f98c61f3d80053b586f82/src/mkdocstrings/templates/python/material/attribute.html#L24-L26

pawamoy:
Yeah just found it, you beat me to it :smile:

pawamoy:
So, since the double tags come from another commit, this PR is OK to be merged!

Issue #178, opened 2020-11-04
Allow using "import" path of an object rather than the "real path" (where it was defined)


pawamoy:
**Is your feature request related to a problem? Please describe.**
Use-cases:
- defining objects in private submodules, importing them in another public facing module.
- importing objects in `__init__.py` modules

Witnessed in #168 and other issues on GitHub.

Since pytkdocs resolves to the real path, these use case are not possible.

**Describe the solution you'd like**
Allow documenting objects from any path.

**Additional context**
pytkdocs will still need to resolve the real path for attributes because it needs to parse their code to find docstrings and type annotations. The real path could be kept in an additional data field for these cases.

pawamoy:
Feature request on pytkdocs: https://github.com/pawamoy/pytkdocs/issues/74

oprypin:
Somewhat related thing that I first brought up in chat. For Python, as you're describing here, an item could be available under multiple import paths, whereas in Crystal there is no namespacing per-file, but there's a feature to explicitly add an alias of a type under another name.

> I'm currently running into an issue where an object can be resolved under more than one name (has an alias), however, the cross-reference-fixer cannot pick up on that, it only knows about the final HTML id.
>
> I understand why that happens, but my wish is that the xref fixer would also ask the collector whether it has a canonical id for this item and look for that id.
>
> Seems like it would have to be a hack like... after trying `url_map` it would also do `for handler in handlers_cache.items(): url_map[handler.canonical_url(identifier)]`

pawamoy:
Well, I'm always in favor of something that makes the code less Python-specific and more language-agnostic, so why not having this canonical ID thing :slightly_smiling_face:

corykinney:
I've been trying to figure out how to have code organized in private submodules with a public interface documented and finally found this issue. Is this still something that might get implemented?

pawamoy:
It is, yes, and it requires some changes on pytkdocs side (for which I couldn't dedicate time yet).

wxdrizzle:
Hi, I noticed that this issue was mentioned as a [pytkdocs issue](https://github.com/mkdocstrings/pytkdocs/issues/74) which is still open, while it was "added to 0.11 in Roadmap" of pytkdocs and seems that the newest version is already 0.12. Then, has this issue been solved? Thank you!

pawamoy:
Hello :slightly_smiling_face: 
[...]

PR #217, opened 2021-01-16
feat: Allow linking to an identifier not only by its exact anchor

oprypin:


pawamoy:
...but also by? :smile: 

I'm not sure to get the change here. Basically the URL map is moved into the Handlers instance, with a method to register an anchor, and a method to get an anchor from a collected item, overwritable in each handler. :+1:

Then when fixing cross-references, we search if we have a registered anchor for a given identifier, and if we do not, we try to *collect* the item through each handler to compute the anchor?

I'd like to discuss the case when the anchor is not already registered. The case happens when someone cross-references an object that was not injected in the docs (it was not collected nor rendered). Then how force-collecting it to get its anchor would be useful, as the object is still not rendered in any page? I think I'm missing something :smile:

oprypin:
The slight mistake in your description is the part about "force-collecting". No such thing is happening, and the object's "linkability" is not affected.

The use case for me: an object has its very strict "canonical" anchor but `collect` (`:::`) has the ability to find it through some more "lax" alias, while `[][]` linking doesn't have that ability, it requires the exact spelling. Maybe you also ran into differences with "spelling" with the new_style thing in Python handler.

oprypin:
> the object's "linkability" is not affected.

because even if the collector returns an anchor and that anchor doesn't exist, the behavior will be exactly the same as without this change

pawamoy:
Oh okay I see now. Your change allows to do something like this:

```python
# pkg/a.py
from pkg._b import Kls
```

```python
# pkg/_b.py
class Kls:
    ...
```

```md
Inject object:

::: pkg._b.Kls

Cross-reference it (in markdown directly or in docstrings):

[my kls][pkg.a.Kls]
```

oprypin:
Did a force push because it contained a commit that was merged elsewhere.
It's a no-op, nothing new to review: https://github.com/pawamoy/mkdocstrings/compare/39d77f8..4789950

oprypin:
Oh I think you're right in your comment. So then, it fixes https://github.com/pawamoy/mkdocstrings/issues/178#issuecomment-740768961 :D  I already forgot

pawamoy:
Well I'd say it's a workaround for #178 and not a full fix, for the following reason: if you import a class from a private module in a module above it, it's because you want to make it public there. Therefore you want that class to use its public path as canonical, not its real path, otherwise it defeats a bit the purpose of documenting your API (and it's the current behavior, that I think is wrong, in pytkdocs). Imagine telling users to inherit from `pkg.a.Model`, but having that class rendered in `pkg/_b` with `#pkg._b.Model` as anchor: not great.

Anyway, not much to do with this PR, I'm mainly commenting on #178 :sweat_smile:

src/mkdocstrings/handlers/base.py
```diff
@@ -386,6 +400,28 @@ def __init__(self, config: dict) -> None:
...
+        """
+        Register that an anchor corresponding to an identifier was encountered when rendering the page.
+
+        Arguments:
+            page: The URL of the current page.
+            anchor: The HTML anchor (without '#') as a string
+        """
+        self._url_map[anchor] = f"{page}#{anchor}"
+
+    def __getitem__(self, identifier: str) -> str:
```
pawamoy, line 415: I understand it makes it work transparently in the `fix_ref` methods, but I think it would be more readable with a real name like `get_url`. Otherwise it's a bit weird to get an URL when doing `handlers.get("identifier")`, especially when the documentation of the Handlers class is "A collection of handlers" (not a collection of URLs or something :smile_cat: ).

oprypin: Definitely agree, was being lazy (minimizing the code diff)

src/mkdocstrings/handlers/base.py
```diff
@@ -386,6 +400,28 @@ def __init__(self, config: dict) -> None:
...
+        try:
+            return self._url_map[identifier]
+        except KeyError:
+            for handler in self._handlers.values():
+                try:
+                    return self._url_map[handler.renderer.get_anchor(handler.collector.collect(identifier, {}))]
```
pawamoy, line 421: What happens if two objects from two files in different languages have the same identifier?

oprypin: As written here, quite simply, whichever handler was ever mentioned first gets priority. The same problem could exist even in the previous state (equal canonical identifiers + whichever identifier got mentioned last), but now a clash becomes a little more likely, but only in the case where neither language's canonical identifier was matched, but both language's "lax" lookup matched.

It seems not so horrible, but can warrant discussion.

pawamoy: Hmmm I guess this is a general issue within projects injecting documentation for two or more languages.

pawamoy: The cross-reference features expects all identifiers and their aliases to be unique. Maybe it's just not powerful enough yet.

pawamoy: Yup. Maybe we could support a handler prefix in the identifier. And use the default handler only when there's no prefix.

src/mkdocstrings/handlers/base.py
oprypin: Pushed a commit to address this

oprypin:
Regressions run shows nothing new https://github.com/oprypin/mkdocstrings-regressions/actions/runs/491976653

pawamoy:
OK, LGTM! Would you also like to update the docs to document this new feature? If not, I can do it of course, but not sure *when*!

oprypin:
OK, added the docs :D
No excuse

pawamoy:
Passing a callable, heh, OK with me :slightly_smiling_face: 
Thank you for the docs, and the typo fixes :smile:

PR #218, opened 2021-01-16
feat: Let handlers add CSS to the pages, do so for Python handler

### LINKED ISSUE #189: Incorporate the recommended CSS style in the package
It's a bit annoying to copy over the recommended CSS in every project. Also asked here: https://github.com/pawamoy/mkdocstrings/pull/159#pullrequestreview-506099310

We can package the recommended stylesheets.
The question is: how to use them in the MkDocs `extra_css` option.

oprypin:
Fixes #189

### DIFF
--- a/src/mkdocstrings/handlers/base.py
+++ b/src/mkdocstrings/handlers/base.py
@@ -15,7 +15,7 @@ import re
 import textwrap
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import Any, Dict, Optional, Sequence
+from typing import Any, Dict, Iterable, Optional, Sequence
 from xml.etree.ElementTree import Element  # noqa: S405 (we choose to trust the XML input)
 
 from jinja2 import Environment, FileSystemLoader
@@ -161,10 +161,12 @@ class BaseRenderer(ABC):
     You can also override the `update_env` method, to add more filters to the Jinja environment,
     making them available in your Jinja templates.
 
-    To define a fallback theme, add a `FALLBACK_THEME` class-variable.
+    To define a fallback theme, add a `fallback_theme` class-variable.
+    To add custom CSS, add an `extra_css` variable or create an 'style.css' file beside the templates.
     """
 
     fallback_theme: str = ""
+    extra_css = ""
 
     def __init__(self, directory: str, theme: str, custom_templates: Optional[str] = None) -> None:
         """
@@ -180,15 +182,21 @@ class BaseRenderer(ABC):
         """
         paths = []
 
-        if custom_templates is not None:
-            paths.append(Path(custom_templates) / directory / theme)
-
         themes_dir = TEMPLATES_DIR / directory
 
         paths.append(themes_dir / theme)
 
-        if self.fallback_theme != "":
+        if self.fallback_theme:
             paths.append(themes_dir / self.fallback_theme)
+
+        for path in paths:
+            css_path = path / "style.css"
+            if css_path.is_file():
+                self.extra_css += "\n" + css_path.read_text(encoding="utf-8")
+                break
+
+        if custom_templates is not None:
+            paths.insert(0, Path(custom_templates) / directory / theme)
 
         self.env = Environment(
             autoescape=True,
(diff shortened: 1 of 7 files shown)


pawamoy:
OK I read your code and it's good, there are just a few things that bother me:

- do you think users will want to hardcode CSS directly in the `extra_css` attribute? Personally I don't think so, because either you write your handler+renderer in a Python package, with templates and therefore you can write CSS in the `style.css`, or you write your handler+renderer in a local Python module (per project), and therefore you can simply write your CSS file as an asset and add it to mkdocs.yml.
- the current approach makes it mandatory for any theme to redeclare the full CSS even if it falls back to another theme which already handles most of the styling. Couldn't we include in the final site both CSS files, so that a theme only has to declare styling for the elements it overwrote?

Based on these two points: couldn't we "simply" copy over the CSS file of the chosen theme (if any) and the CSS file of its fallback theme (if any) to the final site, without reading them into an `extra_css` variable?

oprypin:
> do you think users will want to hardcode CSS directly in the `extra_css` attribute?

Perhaps not, but maybe? Regardless, it's a necessary implementation detail (more at the bottom).
FWIW it's not "hardcode", it's very much possible to generate CSS dynamically in code, just that we're not using this here.

> the current approach makes it mandatory for any theme to redeclare the full CSS even if it falls back to another theme

Is it a real problem to duplicate a little CSS? Conversely, your suggestion I think has a problem.
I don't think the "fallback theme" can generically qualify for being the source of CSS for every theme because, after all, it's also just a theme on its own. So one would be scared to add CSS to that theme just because it might affect other themes?
When overwriting HTML, you overwrite whole files, but here it doesn't work like that.
I also thought generally, how to eliminate this duplication. I didn't find any good way, and decided that it's not worth it.

> couldn't we "simply" copy over the CSS file of the chosen theme (if any) and the CSS file of its fallback theme (if any) to the final site, without reading them into an `extra_css` variable

Now we get to the implementation details.

1. I implemented this generically, so that there can be any number (0 to N) of active handlers. Which (or even how many) handlers got pulled in is not known until all Markdown is finished rendering.
2. It is not possible to change the declaration of which static files get included into the site prior to rendering all the pages (incl. Markdown).
3. It *is* possible to change the *content* of files very late in the build process.
4. Following from these, the best we can do is always specify exactly 1 file to declare. In the end it will contain a merger of all CSS specified by all handlers.
5. The most convenient place to determine which CSS file / content a handler should read is inside its `__init__` because all the path calculations are already there. Having done that, we just store it for later, because at that time / location there isn't anything that has access to write a file (or at least doing it inside `__init__` would be weird).

src/mkdocstrings/handlers/base.py
```diff
@@ -158,16 +160,23 @@ def __init__(self, directory: str, theme: str, custom_templates: Optional[str] =
...
 
-        if self.fallback_theme != "":
+        if self.fallback_theme:
             paths.append(themes_dir / self.fallback_theme)
 
+        for path in paths:
+            css_path = path / "style.css"
+            if css_path.is_file():
+                self.extra_css += "\n" + css_path.read_text(encoding="utf-8")
+                print('!!!', self.extra_css)
```
pawamoy, line 174: debug line :) ?

oprypin: :thinking: sorry, dropped that now

Was double checking everything :smile:

src/mkdocstrings/plugin.py
```diff
@@ -267,9 +273,13 @@ def on_post_build(self, **kwargs) -> None:  # noqa: W0613,R0201 (unused argument
         this hook.
 
         Arguments:
+            config: The MkDocs config object.
             kwargs: Additional arguments passed by MkDocs.
         """
         if self._handlers:
+            css_content = "\n".join(handler.renderer.extra_css for handler in self.handlers.seen_handlers)
+            write_file(css_content.encode("utf-8"), os.path.join(config["site_dir"], self.css_filename))
```
pawamoy, line 281: So I believe this file will be written in the `site` directory by default? Then shouldn't `css_filename` be `assets/stylesheets/_mkdocstrings.css"` instead?

oprypin: Yes.

There isn't any definitively correct path to use, because the paths are based entirely on each theme's own preference. [Material theme](https://github.com/squidfunk/mkdocs-material/tree/master/material) happens to use the path `assets/stylesheets`, while [MkDocs' own themes](https://github.com/mkdocs/mkdocs/tree/master/mkdocs/themes/readthedocs) don't use `assets/`anything at all.

I figured that using just "assets" still achieves the goal of reducing clutter in the output for Material theme, while also not creating the whole `assets/stylesheets` structure just to house this single file when using any other theme.

I don't have a strong preference, you can just tell me any final decision.

pawamoy: Ah yeah I forgot about the themes. Then `assets/_mkdocstrings.css` is perfectly fine. Thanks!

pawamoy:
Alright, not a fan of this string concatenation thing but it's a nitpick and I don't want to prevent a merge just for this :smile: 
Could you add a small paragraph in the docs? Maybe under `handlers/overview/#code`, telling that users can add CSS in the `extra_css` variable or in a `style.css` file? You don't have to be exhaustive, I'd just like to build the docs skeleton along with the features, so it can be improved later :slightly_smiling_face:

oprypin:
Hmm I think documentation of per-theme templates is missing, and this CSS stuff is based on it, so I added both.

PR #219, opened 2021-01-17
refactor: Take anchors from headings directly, don't scan ToC

oprypin:
This has an advantage that random unrelated anchors don't become elibigle for accidental autorefs.
And I have a use case where I want to exclude some items from ToC but stll keep them linkable.

### DIFF
--- a/src/mkdocstrings/handlers/base.py
+++ b/src/mkdocstrings/handlers/base.py
@@ -390,6 +390,8 @@ class Handlers:
     this for the purpose of caching. Use [mkdocstrings.plugin.MkdocstringsPlugin.get_handler][] for convenient access.
     """
 
+    current_page: Optional[str] = None
+
     def __init__(self, config: dict) -> None:
         """
         Initialize the object.
--- a/src/mkdocstrings/extension.py
+++ b/src/mkdocstrings/extension.py
@@ -119,6 +119,9 @@ class AutoDocProcessor(BlockProcessor):
             # So we need to duplicate the headings directly (and delete later), just so 'toc' can pick them up.
             el.extend(headings)
 
+            for heading in headings:
+                self._handlers.register_anchor(self._handlers.current_page, heading.attrib["id"])
+
             parent.append(el)
 
         if the_rest:
--- a/src/mkdocstrings/plugin.py
+++ b/src/mkdocstrings/plugin.py
@@ -31,7 +31,6 @@ from mkdocs.config import Config
 from mkdocs.config.config_options import Type as MkType
 from mkdocs.plugins import BasePlugin
 from mkdocs.structure.pages import Page
-from mkdocs.structure.toc import AnchorLink
 
 from mkdocstrings.extension import MkdocstringsExtension
 from mkdocstrings.handlers.base import BaseHandler, Handlers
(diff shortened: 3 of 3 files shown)


oprypin:
Currently depends on #217, so see only last commit

src/mkdocstrings/handlers/base.py
```diff
@@ -390,6 +390,8 @@ class Handlers:
     this for the purpose of caching. Use [mkdocstrings.plugin.MkdocstringsPlugin.get_handler][] for convenient access.
     """
 
+    current_page: Optional[str] = None
```
oprypin, line 393: Honestly at this point I'm really abusing the fact that `Handlers` happens to be available to both the plugin and the extension. Maybe later I'll see a better way but right now I don't see.

pawamoy: Don't worry, putting data in-between the plugin and the extension was the primary design :slightly_smiling_face: It's impossible without. But yeah, maybe later, with more hindsight we will see a way to refactor elegantly.

oprypin:
Regressions run shows 2 changed repos.
E.g. https://github.com/oprypin/mkdocstrings-regressions/runs/1717561357?check_suite_focus=true
They were "abusing" the ability to link any heading (not just mkdocstrings identifiers) through mkdocstrings autoref. I'm intentionally removing that ability, so as for me this is fine. Let me know if you disagree.

pawamoy:
Oh well, this feature is clearly advertised in the README, so I'm not sure we should remove it :confused:

> Cross-references to other objects: mkdocstrings makes it possible to reference other headings from your Markdown files with the classic Markdown syntax: [identifier][] or [title][identifier]. This feature is language agnostic as well: you can cross-reference any heading that appear in your Markdown pages. If the handler for a particular language renders headings for documented objects, you'll be able to reference them!

oprypin:
Oh, I see :(

Myself, not knowing about this ability, I actually ran into the issue that it's not feasible to make a real link to a normal doc page from inside a docstring, because "relativity" of the URL is not possible to know in advance.

So I suppose the need for this is there.

But I think I don't like this particular way of solving the need, perhaps it's too much against the spirit of Markdown. Would've liked a more direct solution. And, of course, it's weird that it'll probably pick an arbitrary page if multiple pages happen to have the same heading.

pawamoy:
I'm looking at plugins that allow to do that, to see if we could remove the feature from mkdocstrings. [This plugin](https://github.com/Jackiexiao/mkdocs-roamlinks-plugin) seems to do the trick, although the syntax changes a bit: `[title][id]` becomes `[[title]]`.

oprypin:
https://github.com/mkdocs/mkdocs/wiki/MkDocs-Plugins#autoreflinks would probably be the most direct replacement.
`[Title]`
But [the implementation](https://github.com/pauloue/mkdocs-autoreflinks-plugin/blob/master/autoreflinks/plugin.py) is BeautifulSoup

https://github.com/midnightprioriem/mkdocs-autolinks-plugin itself is a bit less intrusive, I think.
`[Text](page.md#id)` (resolves a document that exists anywhere)

https://github.com/sander76/mkdocs-abs-rel-plugin is the least intrusive while still solving the problem (the problem as I saw it, anyway).
`[Text](/abs/path/page.md#id)`

**But** the latter two only look at `page_markdown`, meaning that they will never get to see our inserted Markdown sub-documents.

The outlook of the available implementations is quite bad, it seems.

pawamoy:
Oh, good catch about `page_markdown` being too early for us.

oprypin:
(The force push is no-op, purely rebasing)

oprypin:
So there wasn't any good autorefs plugin, huh? Well, now there is :D #220

Let me know what you think. Naming and other technicalities are up for debate, of course.

oprypin:
Now definitely closing this in favor of https://github.com/pawamoy/mkdocstrings/pull/220

PR #220, opened 2021-01-20
refactor: Split out autorefs plugin, make it optional

oprypin:
"autorefs" is now a self-sufficient MkDocs plugin, just allowing linking to headings on any page.

As for mkdocstrings, unless users opt into using autorefs for any heading (done by scanning the ToC as before), this will now only do autorefs to mkdocstrings identifiers.
Opting in is done by just enabling 'autorefs' as a separate plugin, which it now is.

The new behavior of linking only to identifiers is implemented by force-including the autorefs plugin even if it's not specified, but in a "subdued" mode. Idea inspired by the interaction between pymdownx.{superfences,highlight}.


pawamoy:
A rebase and some quality fixes (missing docstrings mostly), and we're good to go.

pawamoy:
OK, the tornado vulnerability can be fixed later with a dependency upgrade, lets' merge!

PR #229, opened 2021-02-09
Migrate markdown_include -> pymdownx.snippets

oprypin:


pawamoy:
What's up with Poetry :thinking:

oprypin:
It's the same kind of situation - "this plugin but actually working"

Or, well, apparently "markdown-include" has a great feature to inherit Markdown heading levels, but it's almost never needed.

pawamoy:
OK, good to me then, one dependency less.

PR #232, opened 2021-02-10
chore: Template upgrade

pawamoy:
- Don't autoescape when rendering changelog
- Generate credits on-the-fly, when building docs
  (for this we're using mkdocs-macros-plugins)
- Update docs deploy action to point to org pages
- Add cleancov argument to test action,
  to allow combining coverage for all tested Python versions
- Reorganize docs navigation
- Add coverage HTML report into docs
- Update dev-dependencies, remove ipython


docs/macros.py
```diff
@@ -0,0 +1,77 @@
...
+    when building the documentation.
+
+    Arguments:
+        env: An object used to add macros and filters to the environment.
+    """
+
+    @env.macro  # noqa: WPS430 (nested function)
+    def credits():  # noqa: W0612,W0622,WPS430 (unused, shadows credits)
+        jinja_env = SandboxedEnvironment(undefined=StrictUndefined)
+        template_url = "https://raw.githubusercontent.com/pawamoy/jinja-templates/master/credits.md"
```
oprypin, line 74: Mm I don't want to be executing arbitrary code when generating the docs. Easiest thing to do is to specify the commit hash, not `master`.

mkdocs.yml
```diff
@@ -21,16 +26,15 @@ nav:
...
-- License: license.md
+- Development:
+  - Contributing: contributing.md
+  - Code of Conduct: code_of_conduct.md
+  - Coverage report: coverage.md
 
 theme:
   name: material
+  features:
+  - navigation.tabs
```
oprypin, line 37: Might as well set up https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#navigation-expansion while we're here

docs/macros.py
```diff
@@ -0,0 +1,77 @@
...
+    for pkg in search_packages_info(dependencies):
+        pkg = {_: pkg[_] for _ in ("name", "home-page")}
+        packages[pkg["name"].lower()] = pkg
+
+    loop = asyncio.get_event_loop()
+    client = httpx.AsyncClient()
+    to_get = [dep for dep in dependencies if dep not in packages]
+    coroutines = [client.get(f"https://pypi.python.org/pypi/{dep}/json") for dep in to_get]
+    responses = loop.run_until_complete(asyncio.gather(*coroutines))
+    loop.run_until_complete(client.aclose())
```
oprypin, line 43: I'm getting `RuntimeError: This event loop is already running` when this is hit for the 2nd time within `mkdocs serve`

docs/macros.py
pawamoy: True, I'll do that, thanks

mkdocs.yml
pawamoy: Indeed, will do, thanks

docs/macros.py
pawamoy: Ha! I missed that. Will run some tests and fix it.

docs/macros.py
```diff
@@ -0,0 +1,77 @@
...
+from pathlib import Path
+
+import httpx
+import toml
+from jinja2 import StrictUndefined
+from jinja2.sandbox import SandboxedEnvironment
+from pip._internal.commands.show import search_packages_info  # noqa: WPS436 (no other way?)
+
+
+def get_credits_data() -> dict:
```
oprypin, line 14: I was thinking, it sucks that we'll have to ping the pypi site on every build. But no easy solution, so whatever.

However, one easy thing to do is `@functools.lru_cache()`, that'll at least make it stick within repeats in `mkdocs serve`

pawamoy: Yeah I was thinking about caching results on the disk. It's really important if the cache is out-of-date.

docs/macros.py
pawamoy: Damn I struggle to find a solution to this.

oprypin: An idea. Maybe drop the download part altogether? There are only like 5 packages that are queried. Either just deal without that information or hardcode it for those packages.

pawamoy: You know what you're right. The packages that pip could not find are the one that were not installed because they're now part of the standard lib or only useful or Windows. Let's drop them. I guess my ideas just need to be confronted from time to time :grinning:

oprypin: Don't forget to drop the httpx dependency at the end 🤭

oprypin (approved):
Good change. Just check what's up with the coverage-related error.

pawamoy:
Yes I already have the fix for it, will push tomorrow.

docs/credits.md
```diff
@@ -1 +1 @@
---8<-- "CREDITS.md"
```
oprypin, line 1: Should the file CREDITS.md be deleted from the repo now?

pawamoy: Yes!

PR #236, opened 2021-02-14
tests: Build test sites into a temp directory

oprypin:
## chore: Migrate from flakehell to flake8
  
  Flakehell is abandoned, I personally ran into problems with its faulty lingering cache, and now it's preventing me from configuring darglint.

## tests: Build test sites into a temp directory

  This was causing random test failures because multiple tests concurrently try to delete and re-create the same site dir.
  Refactor to use pytest fixtures instead while on it.

  Bonus: allow docstrings to not specify Arguments sections etc. But if they are specified, they will still be checked.
  Here because it's ridiculous to document fixtures in every test function. Generally because sometimes all args are obvious if it's just an override.

### DIFF
--- a/config/flake8.ini
+++ b/config/flake8.ini
@@ -1,0 +1,50 @@
+[flake8]
+exclude = fixtures,docs,site
+max-line-length = 132
+ignore =
+    # we write docstrings in markdown, not rst
+    RST*,
+    # redundant with W0622 (builtin override), which is more precise about line number
+    A001,
+    # missing docstring in magic method
+    D105,
+    # multi-line docstring summary should start at the first line
+    D212,
+    # whitespace before ‘:’ (incompatible with Black)
+    E203,
+    # redundant with E0602 (undefined variable)
+    F821,
+    # black already deals with quoting
+    Q000,
+    # use of assert
+    S101,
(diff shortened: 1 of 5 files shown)


config/flake8.ini
```diff
@@ -0,0 +1,50 @@
...
+    F821,
+    # black already deals with quoting
+    Q000,
+    # use of assert
+    S101,
+    # we are not parsing XML
+    S405,
+    # line break before binary operator (incompatible with Black)
+    W503,
+    "-C0103", # two-lowercase-letters variable DO conform to snake_case naming style
```
pawamoy, line 26: Any reason you're adding the comment at the end of this line and not the others?
If possible I'd to have comments on the same lines, even if codes must be wrapped in double-quotes. What do you think?

pawamoy: Also why does it have the `-` prefix and not the others?

oprypin: I... have no idea how this appeared here.

And no, inline comments are not possible.
https://flake8.pycqa.org/en/3.8.4/user/configuration.html#project-configuration

pawamoy (approved):

PR #242, opened 2021-02-24
chore: Automatically generate API reference stub files

oprypin:


pyproject.toml
```diff
@@ -44,6 +44,7 @@ isort = {version = "^5.7.0", extras = ["pyproject"]}
 jinja2-cli = "^0.7.0"
 mkdocs-coverage = "^0.2.1"
 mkdocs-macros-plugin = "^0.5.0"
+mkdocs-gen-files = {version = "^0.3.0", markers = "python_version>='3.7'"}
```
pawamoy, line 47: Does it mean we won't be able to build docs on Python 3.6?

pawamoy (dismissed):
:+1: 

Just one question, see review!

pawamoy:
Maybe we should remove the test that builds the docs. But if we do we should add more unit tests :smile:

mkdocs.yml
```diff
@@ -58,6 +59,9 @@ markdown_extensions:
 
 plugins:
 - search
+- gen-files:
+    scripts:
+    - scripts/gen_doc_stubs.py
 - section-index
```
pawamoy, line 65: Is this still needed?

mkdocs.yml
```diff
@@ -58,6 +59,9 @@ markdown_extensions:
 
 plugins:
 - search
+- gen-files:
+    scripts:
+    - scripts/gen_doc_stubs.py
```
pawamoy, line 64: What do you think about putting the script in the docs directory instead, just like I did for macros.py?

pawamoy (commented):
In fact I have more questions :sweat_smile:

pyproject.toml
oprypin: Yes. I also looked into supporting 3.6 in my plugin but I confirmed that a very important feature is needed from Python 3.7, so it's not viable.

mkdocs.yml
oprypin: Yes, it's 2 different things. Clickable sections in the nav. Build the site and check it out if you haven't lately :sweat_smile:

mkdocs.yml
oprypin: Sure, I always do it that way. I started this commit before you added macros :sweat_smile: 

Note BTW that those 'macros.py' will end up browsable in the final site, but it's OK

pyproject.toml
pawamoy: Out of curiosity, which feature is this :stuck_out_tongue: ?

mkdocs.yml
pawamoy: Ah, I'm dumb, I checked the published docs, which are not up-to-date. Thanks :slightly_smiling_face:

mkdocs.yml
pawamoy: Ah, interesting. But like you said, it's OK.

pyproject.toml
oprypin: Module-level __getattr__
https://www.python.org/dev/peps/pep-0562/
https://github.com/oprypin/mkdocs-gen-files/blob/96cc72d8b84aa0593b782424502541c736dd7e11/mkdocs_gen_files/__init__.py#L18

pawamoy: I see, thanks. It's OK if the docs can't be built using Python 3.6. CI is already using 3.8 anyway.

mkdocs.yml
oprypin: Done (this, and also moved the script)

pawamoy (approved):
Perfect, thanks!

PR #248, opened 2021-02-27
ci: Don't fail when checking docs on Python 3.6

pawamoy:


duties.py
```diff
@@ -201,8 +201,8 @@ def check_docs(ctx):
     Arguments:
         ctx: The context instance (passed automatically).
     """
-    # pytkdocs fails on Python 3.9 for now
-    nofail = sys.version.startswith("3.9")
+    # mkdocs-gen-files works on 3.7+ only
+    nofail = sys.version.startswith("3.6")
```
oprypin, line 205: General comment (and might as well change it while we're here): this is not a safe way to check for versions, as it includes version 3.60 etc.

```suggestion
    nofail = sys.version_info < (3, 7)
```

pawamoy: Oooooh you're right, thanks!

Issue #251, opened 2021-03-10
Inventory Object (like in sphinx)


FalseDev:
**Is your feature request related to a problem? Please describe.**

**Background:** I'm making [a discord bot for developers](https://github.com/FalseDev/Tech-Struck) that has an RTFM command (searches through docs).

**How it works:** [Sphinx](https://www.sphinx-doc.org/) generates an `objects.inv` file which can be fetched and used like an index, I use this to build an index of all documentation  under a module and search through it

**My Problem:** Currently mkdocstring doesn't (as far as I know) produce any similar *index/inventory* object

 **Describe the solution you'd like**

The ideal solution would be making mkdocstrings create an inventory (`objects.inv`) object fully compatible with sphinx

**Describe alternatives you've considered**
I could use web scraping, but it's much more resource intensive and takes longer, so I couldn't use it here

pawamoy:
Hi @FalseDev, thanks for opening this :slightly_smiling_face: We have plans for this, yes! I think it would even make sense to implement it in mkdocstrings first, and then use it in pytkdocs/mkdocstrings for automatic cross-documentation links :slightly_smiling_face:

pawamoy:
Implemented in #253 :slightly_smiling_face:

PR #253, opened 2021-03-10
feat: Generate objects inventory

pawamoy:
Feature request #251

### DIFF
--- a/src/mkdocstrings/handlers/base.py
+++ b/src/mkdocstrings/handlers/base.py
@@ -24,6 +24,7 @@ from mkdocstrings.handlers.rendering import (
     IdPrependingTreeprocessor,
     MkdocstringsInnerExtension,
 )
+from mkdocstrings.inventory import Inventory
 from mkdocstrings.loggers import get_template_logger
 
 CollectorItem = Any
@@ -163,6 +164,7 @@ class BaseRenderer(ABC):
         content: str,
         heading_level: int,
         *,
+        role: Optional[str] = None,
         hidden: bool = False,
         toc_label: Optional[str] = None,
         **attributes: str,
@@ -172,6 +174,7 @@ class BaseRenderer(ABC):
         Arguments:
             content: The HTML within the heading.
             heading_level: The level of heading (e.g. 3 -> `h3`).
+            role: An optional role for the object bound to this heading.
             hidden: If True, only register it for the table of contents, don't render anything.
             toc_label: The title to use in the table of contents ('data-toc-label' attribute).
             attributes: Any extra HTML attributes of the heading.
@@ -184,6 +187,8 @@ class BaseRenderer(ABC):
         if toc_label is None:
             toc_label = content.unescape() if isinstance(el, Markup) else content  # type: ignore
         el.set("data-toc-label", toc_label)
+        if role:
+            el.set("data-role", role)
         self._headings.append(el)
 
         if hidden:
@@ -285,7 +290,12 @@ class BaseHandler:
     Inherit from this class to implement a handler.
 
     It's usually just a combination of a collector and a renderer, but you can make it as complex as you need.
+
+    Attributes:
+        domain: The cross-documentation domain/language for this handler.
     """
+
+    domain: str = "default"
 
     def __init__(self, collector: BaseCollector, renderer: BaseRenderer) -> None:
         """Initialize the object.
@@ -314,6 +324,7 @@ class Handlers:
         """
         self._config = config
         self._handlers: Dict[str, BaseHandler] = {}
+        self.inventory: Inventory = Inventory(project=self._config["site_name"])
 
     def get_anchor(self, identifier: str) -> Optional[str]:
         """Return the canonical HTML anchor for the identifier, if any of the seen handlers can collect it.
--- a/src/mkdocstrings/handlers/base.py
+++ b/src/mkdocstrings/handlers/base.py
@@ -293,9 +293,12 @@ class BaseHandler:
 
     Attributes:
         domain: The cross-documentation domain/language for this handler.
+        enable_inventory: Whether this handler is interested in enabling the creation
+            of the `objects.inv` Sphinx inventory file.
     """
 
     domain: str = "default"
+    enable_inventory: bool = False
 
     def __init__(self, collector: BaseCollector, renderer: BaseRenderer) -> None:
         """Initialize the object.
--- a/src/mkdocstrings/inventory.py
+++ b/src/mkdocstrings/inventory.py
@@ -70,11 +70,8 @@ class Inventory(dict):
         item = InventoryItem(*args, **kwargs)
         self[item.name] = item
 
-    def format_sphinx(self, compress: bool = True) -> bytes:
+    def format_sphinx(self) -> bytes:
         """Format this inventory as a Sphinx `objects.inv` file.
-
-        Arguments:
-            compress: Whether to compress the data using zlib.
 
         Returns:
             The inventory as bytes.
@@ -96,7 +93,4 @@ class Inventory(dict):
         for item in self.values():
             lines.append(item.format_sphinx().encode("utf8"))
 
-        data = b"\n".join(lines)
-        if compress:
-            data = zlib.compress(data, 9)
-        return header + data
+        return header + zlib.compress(b"\n".join(lines), 0)
--- a/src/mkdocstrings/inventory.py
+++ b/src/mkdocstrings/inventory.py
@@ -89,8 +89,6 @@ class Inventory(dict):
             .encode("utf8")
         )
 
-        lines = []
-        for item in self.values():
-            lines.append(item.format_sphinx().encode("utf8"))
-
+        lines = [item.format_sphinx().encode("utf8") for item in self.values()]
+        # we use compression level 0 to make the file readable by humans
         return header + zlib.compress(b"\n".join(lines), 0)
--- a/src/mkdocstrings/inventory.py
+++ b/src/mkdocstrings/inventory.py
@@ -1,0 +1,102 @@
+"""Module responsible for the objects inventory."""
+
+# Credits to Brian Skinn and the sphobjinv project:
+# https://github.com/bskinn/sphobjinv
+
+import zlib
+from textwrap import dedent
+from typing import List, Optional
+
+
+class InventoryItem:
+    """Inventory item."""
+
+    def __init__(self, name: str, domain: str, role: str, uri: str, priority: str = "1", dispname: str = "-"):
+        """Initialize the object.
+
+        Arguments:
+            name: The item name.
+            domain: The item domain, like 'python' or 'crystal'.
+            role: The item role, like 'class' or 'method'.
+            uri: The item URI.
+            priority: The item priority. It can help for inventory suggestions.
+            dispname: The item display name.
+        """
+        self.name: str = name
+        self.domain: str = domain
+        self.role: str = role
+        self.uri: str = uri
+        self.priority: str = priority
+        self.dispname: str = dispname
+
+    def format_sphinx(self) -> str:
+        """Format this item as a Sphinx inventory line.
+
+        Returns:
+            A line formatted for an `objects.inv` file.
+        """
+        uri = self.uri
+        name_length = len(self.name)
+        if uri[-name_length - 1 :] == "#" + self.name:
+            uri = uri[:-name_length] + "$"
+        return f"{self.name} {self.domain}:{self.role} {self.priority} {uri} {self.dispname}"
+
+
+class Inventory(dict):
+    """Inventory of collected and rendered objects."""
+
+    def __init__(self, items: Optional[List[InventoryItem]] = None, project: str = "project", version: str = "0.0.0"):
+        """Initialize the object.
+
+        Arguments:
+            items: A list of items.
+            project: The project name.
+            version: The project version.
+        """
+        super().__init__()
+        items = items or []
+        for item in items:
+            self[item.name] = item
+        self.project = project
+        self.version = version
+
+    def register(self, *args, **kwargs):
+        """Create and register an item.
+
+        Arguments:
+            *args: Arguments passed to [InventoryItem][mkdocstrings.inventory.InventoryItem].
+            **kwargs: Keyword arguments passed to [InventoryItem][mkdocstrings.inventory.InventoryItem].
+        """
+        item = InventoryItem(*args, **kwargs)
+        self[item.name] = item
+
+    def format_sphinx(self, compress: bool = True) -> bytes:
+        """Format this inventory as a Sphinx `objects.inv` file.
+
+        Arguments:
+            compress: Whether to compress the data using zlib.
+
+        Returns:
+            The inventory as bytes.
+        """
+        header = (
+            dedent(
+                f"""
+                # Sphinx inventory version 2
+                # Project: {self.project}
+                # Version: {self.version}
+                # The remainder of this file is compressed using zlib.
+                """
+            )
+            .lstrip()
+            .encode("utf8")
+        )
+
+        lines = []
+        for item in self.values():
+            lines.append(item.format_sphinx().encode("utf8"))
+
+        data = b"\n".join(lines)
+        if compress:
+            data = zlib.compress(data, 9)
+        return header + data
--- a/src/mkdocstrings/plugin.py
+++ b/src/mkdocstrings/plugin.py
@@ -156,6 +156,7 @@ class MkdocstringsPlugin(BasePlugin):
             theme_name = config["theme"].name
 
         extension_config = {
+            "site_name": config["site_name"],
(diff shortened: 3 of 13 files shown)


FalseDev (commented):
Seeing from the Github Action logs, dependencies will have to be updated to include the `sphobjinv` module.

pyproject.toml
```diff
@@ -27,6 +27,7 @@ mkdocs = "^1.1"
 mkdocs-autorefs = "^0.1"
 pymdown-extensions = ">=6.3, <9.0"
 pytkdocs = ">=0.2.0, <0.12.0"
+sphobjinv = "^2.0.1"
```
oprypin, line 30: Not an acceptable dependency to always include, in my view. Should be optional.

https://github.com/bskinn/sphobjinv/blob/b8dbe8e975e502bac065519d82a103ea6435aa5b/setup.cfg#L37-L40

pawamoy: I think we can even get rid of it. Looking at the uncompressed objects.inv, it seems to be easy to generate:

```
# Sphinx inventory version 2
# Project: project
# Version: 0.0.0
# The remainder of this file is compressed using zlib.
mkdocstrings.extension py:  reference/extension/#mkdocstrings.extension 
mkdocstrings.extension.AutoDocProcessor py:  reference/extension/#mkdocstrings.extension.AutoDocProcessor 
mkdocstrings.extension.AutoDocProcessor.__init__ py:  reference/extension/#mkdocstrings.extension.AutoDocProcessor.__init__ 
mkdocstrings.extension.AutoDocProcessor.run py:  reference/extension/#mkdocstrings.extension.AutoDocProcessor.run 
mkdocstrings.extension.AutoDocProcessor.test py:  reference/extension/#mkdocstrings.extension.AutoDocProcessor.test 
mkdocstrings.extension.MkdocstringsExtension py:  reference/extension/#mkdocstrings.extension.MkdocstringsExtension
```

bskinn: Author of `sphobjinv` here. Yep, it's pretty easy to construct manually. The only really tricky part is the partial zlib compression of only the objects data.

It looks like you're missing a number of elements in the object lines of that inventory; specifically `role`, `priority` and `dispname`. See the syntax for an uncompressed inventory [here](https://sphobjinv.readthedocs.io/en/stable/syntax.html).

Also, FWIW `sphobjinv`'s implementation of the partial zlib compression is [here](https://github.com/bskinn/sphobjinv/blob/260a96cd4eaa7d5cdd8e12f03ae7376243841fa6/src/sphobjinv/zlib.py#L96-L139).

pawamoy: Hey @bskinn, thanks a lot for chiming in!

Noted for the zlib partial compression, thanks.

About the elements missing: I guess I can rework the PR so that 1) we don't rely on autorefs to get the registered objects and 2) we also store each objects "type" (class, method), though it depends on the language, so not sure yet how to do this. Then I'll be able to set the role. I can also default the priority to 1. I see that I can also shrink the contents with `$` instead of the name. And add `-` at the end.

src/mkdocstrings/inventory.py
```diff
@@ -7,7 +7,8 @@ def get_inventory(autorefs, project="project", version="0.0.0"):
     inv.version = version
 
     for anchor, url in autorefs._url_map.items():
-        o = soi.DataObjStr(name=anchor, domain="py", uri=url, role="", priority="", dispname="")
+        url = url.replace(anchor, "$")
```
bskinn, line 10: Minor potential gotcha here -- this `$` shorthand will only be interpreted correctly by Sphinx when it replaces `anchor` at the *very end* of `url`.

If you know that that'll always be the case, then you should be fine.

pawamoy: Yes it's not the most robust code right now, we'll to be sure of that indeed :slightly_smiling_face:

src/mkdocstrings/inventory.py
```diff
@@ -0,0 +1,22 @@
...
+
+    lines = []
+    for anchor, url in autorefs._url_map.items():
+        url = url.replace(anchor, "$")
+        lines.append(f"{anchor} py:obj 1 {url} -".encode("utf8"))
+
+    data = b"\n".join(lines)
+    if compress:
+        data = zlib.compress(data, 9)
+    return header + data
```
pawamoy, line 22: So I've just implemented this version that doesn't use sphobjinv, and it seems to work fine.

bskinn: &lt;nod&gt;, looks pretty good to me.

If you don't mind adding Sphinx to your dev/test dependencies, there's a not-horribly-painful way to sanity-check whether Sphinx can read a generated inventory; see e.g. [here](https://github.com/bskinn/sphobjinv/blob/260a96cd4eaa7d5cdd8e12f03ae7376243841fa6/conftest.py#L163-L177).

pawamoy: Sure, will do, too bad Sphinx itself is so heavy :sweat_smile: 

```
  • Installing pytz (2021.1)
  • Installing alabaster (0.7.12)
  • Installing babel (2.9.0)
  • Installing docutils (0.16)
  • Installing imagesize (1.2.0)
  • Installing requests (2.15.1)
  • Installing sphinxcontrib-applehelp (1.0.2)
  • Installing sphinxcontrib-devhelp (1.0.2)
  • Installing sphinxcontrib-htmlhelp (1.0.3)
  • Installing sphinxcontrib-jsmath (1.0.1)
  • Installing sphinxcontrib-qthelp (1.0.3)
  • Installing sphinxcontrib-serializinghtml (1.1.4)
  • Installing sphinx (3.5.2)
```

If you have any idea on how I could avoid all these contribs in the dev-deps, I'm all ears :slightly_smiling_face:

bskinn: @pawamoy In a quick test, I was able to use `sphinx.util.inventory.InventoryFile` successfully with just a `pip install sphinx docutils --no-deps`:

```
>>> from sphinx.util.inventory import InventoryFile
>>> with open("/path/to/objects_attrs.inv", "rb") as f:
...   InventoryFile.load(f, "", os.path.join)
...
{'py:class': {'attr.Attribute':  ... }}
```

pawamoy: Nice! I'll see if I can set this pip install command as a task or something, as I won't be able to prevent Poetry from installing subdeps.

bskinn: Yep, should be possible to have that extra `pip install` just be an additional `run` step in your workflow. I do it all the time.

src/mkdocstrings/inventory.py
pawamoy: I'm now replacing only if there's a `#` before the anchor, and the anchor is at the end.

pawamoy:
So here's the approach I took, ending up using `do_heading` like you suggested @oprypin:

- implemented `Inventory` and `InventoryItem` classes, nothing crazy here
- an inventory item has a name, a domain, a role, an URI, a priority and a display name, just like Sphinx's inventories
- upon rendering an object's title in a template, we pass an additional `role` parameter to the `do_heading` special filter
- `do_heading` renders that title to HTML and sets a `data-role` attribute if a role was given
- it allows us to get it back in the main extension's `run` method
- `_process_block` returns the handler directly instead of its renderer collected headings, so we can both get the headings in the `run` method as well as the domain on the handler (which is `"py"` on the python handler)
- we have everything we need to then register that heading/object in the inventory instance that is attached to the `Handlers` unique instance
- finally, in the `post_build` hook, we simply format that inventory to Sphinx's objects.inv and add it to the site files

This implementation does not register headings that are not objects. It will also allow extensibility if we ever want to generate inventories in other formats. I reused Sphinx's attributes for an inventory item (name, domain, role, etc.) because I think it makes sense in the context of a language-agnostic documentation tool.

If we go ahead with this, it means you'll have to update your crystal handler @oprypin, to add a `domain` attribute to the handler class (`cr`? `crystal`?), and to pass roles from the templates (not sure what roles we can have in crystal).

This PR is out of draft-mode, so don't hesitate to review it thoroughly, I'm open to criticism :slightly_smiling_face:

src/mkdocstrings/templates/python/material/attribute.html
```diff
@@ -17,6 +17,7 @@
       {% endif %}
 
       {% filter heading(heading_level,
+          role="data" if obj == module else "attr",
```
pawamoy, line 20: I must add tests for this. Not sure if it would fail silently or crash when someone renders `::: something.that.isnot.a.module` and therefore the `module` variable does not exist.

src/mkdocstrings/handlers/base.py
```diff
@@ -314,6 +324,7 @@ def __init__(self, config: dict) -> None:
         """
         self._config = config
         self._handlers: Dict[str, BaseHandler] = {}
+        self.inventory: Inventory = Inventory()
```
pawamoy, line 327: Project and version should be set here when instantiating the inventory.

src/mkdocstrings/inventory.py
```diff
@@ -0,0 +1,99 @@
...
+
+class Inventory(dict):
+    """Inventory of collected and rendered objects."""
+
+    HEADER = """
+        # Sphinx inventory version 2
+        # Project: {project}
+        # Version: {version}
+        # The remainder of this file is compressed using zlib.
+    """
```
pawamoy, line 54: Move this into the `format_sphinx` method.

src/mkdocstrings/handlers/base.py
pawamoy: We'll only set the project name for now, using `site_name`. We'll see later for the version.

FalseDev:
Is there any updates on this pr?

pawamoy:
Sure! I worked on this yesterday, let me push some changes :slightly_smiling_face:

pawamoy:
Just some tests to write and it'll be good to merge :slightly_smiling_face:

tests/test_inventory.py
```diff
@@ -0,0 +1,40 @@
...
+try:
+    from sphinx.util.inventory import InventoryFile
+except ImportError:
+    InventoryFile = None  # type: ignore
+
+MKDOCSTRINGS_OBJECTS_INV = Path("site/objects.inv")
+
+
+# https://github.com/bskinn/sphobjinv/blob/260a96cd4eaa7d5cdd8e12f03ae7376243841fa6/conftest.py#L163-L177
+@pytest.mark.skipif(InventoryFile is None, reason="Sphinx is not installed")
```
bskinn, line 20: You probably want this `skipif` on all the tests in the module, right?

I usually use an `autouse` fixture to skip a whole test module based on a condition, e.g., [here](https://github.com/bskinn/sphobjinv/blob/master/tests/test_api_good_nonlocal.py#L36-L44).

Something to maybe consider.

tests/test_inventory.py
```diff
@@ -0,0 +1,40 @@
...
+    [
+        Inventory(),
+        Inventory([InventoryItem(name="object_path", domain="py", role="obj", uri="page_url")]),
+        Inventory([InventoryItem(name="object_path", domain="py", role="obj", uri="page_url#object_path")]),
+        Inventory([InventoryItem(name="object_path", domain="py", role="obj", uri="page_url#other_anchor")]),
+    ],
+)
+def test_sphinx_load_inventory_file(inventory):
+    """Perform the 'live' inventory load test."""
+    buffer = BytesIO(inventory.format_sphinx())
```
bskinn, line 32: I'm totally going to steal this for my own tests, avoid a bunch of unnecessary disk interactions. Kicking myself for not thinking of it myself. :-)

tests/test_inventory.py
pawamoy: Indeed, thanks!

tests/test_inventory.py
pawamoy: Hehehehe please do :smile:

src/mkdocstrings/templates/python/material/attribute.html
pawamoy: It works fine with module-level and class-level attributes, in both cases (rendering a module recursively or just a class recursively), so I'll go ahead without tests. But it would be nice to have some end-to-end rendering tests at some point.

src/mkdocstrings/plugin.py
```diff
@@ -206,6 +207,9 @@ def on_post_build(self, config: Config, **kwargs) -> None:  # noqa: W0613,R0201
             log.debug("Tearing handlers down")
             self._handlers.teardown()
 
+            inv_contents = self._handlers.inventory.format_sphinx()
+            write_file(inv_contents, os.path.join(config["site_dir"], "objects.inv"))
```
oprypin, line 211: Wait,.. shouldn't adding this file be optional? For Crystal people will never use this but they will be annoyed by the file

tests/test_inventory.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@pytest.fixture(scope="module", autouse=True)
+def _skip_if_sphinx_missing():
+    """Skip test if Sphinx is not installed.
+
+    Auto-applied to all functions in module.
+    """
+    if InventoryFile is None:
+        pytest.skip("Sphinx is not installed")
```
oprypin, line 26: I think you can just write `inventory = pytest.importorskip('sphinx.util.inventory')` and you don't need any of the explicit `skip` stuff.

https://docs.pytest.org/en/latest/how-to/skipping.html#skipping-on-a-missing-import-dependency

src/mkdocstrings/plugin.py
pawamoy: That's why I needed your review :smile: For the obvious things I missed! You're right, this should be made optional. Or we could generate it only if the Python handler was used.

tests/test_inventory.py
pawamoy: Ah, nice, 1 line instead of 12, I'll try that, thanks.

src/mkdocstrings/plugin.py
oprypin: Ah so now I understand that adding these files by default is a good goal. I would suggest yeah, making it configurable, and enabled by default only for Python handler. But the config is still that all handlers produce inventory or none do, just that the default happens to be True only if there is at least 1 handler that asks for inventory. Kinda pseudo code:

```
try:
    use_inventory = self.config["enable_inventory"]
except KeyError:
    use_inventory = any(getattr(h, "wants_inventory", False) for h in handlers)
if use_inventory:
    write_file......
```

src/mkdocstrings/inventory.py
```diff
@@ -0,0 +1,102 @@
...
+            .encode("utf8")
+        )
+
+        lines = []
+        for item in self.values():
+            lines.append(item.format_sphinx().encode("utf8"))
+
+        data = b"\n".join(lines)
+        if compress:
+            data = zlib.compress(data, 9)
```
oprypin, line 101: Oh I just remembered that the files aren't human-readable at all.
But I sniped the fact that you actually don't have to compress them. 
Like here: https://docs.python.org/2.6/objects.inv
I think the times are different now, compression is kinda handled transparently everywhere, and mkdocs isn't very good for huge sites anyway. So let's just not compress it maybe?

oprypin: Aha we just found that "version 1" doesn't *only* mean the lack of compression but also other bad changes.
But...... you can have a kinda-uncompressed version 2.

```
# Sphinx inventory version 2
# Project: mkdocstrings
# Version: 0.0.0
# The remainder of this file is compressed using zlib.
x^A^A"%<DD><DA>mkdocstrings.extension py:mod 1 reference/extension/#$ -
mkdocstrings.extension.AutoDocProcessor py:class 1 reference/extension/#$ -
mkdocstrings.extension.AutoDocProcessor.__init__ py:meth 1 reference/extension/#$ -
...
```
`data = zlib.compress(data, 0)` 😳

bskinn:
Yeah, 'version 1' is missing important fields, IIRC.

That's....hackety-wow, geez. 😝

src/mkdocstrings/handlers/base.py
```diff
@@ -184,6 +187,8 @@ def do_heading(
         if toc_label is None:
             toc_label = content.unescape() if isinstance(el, Markup) else content  # type: ignore
         el.set("data-toc-label", toc_label)
+        if role:
+            el.set("data-role", role)
```
oprypin, line 191: Note that the "data-role" never ends up in any HTML, and this information could've been associated with the item in any other way. But yeah this is the most straightforward way that we have now.

oprypin (approved):
Nice!

src/mkdocstrings/handlers/base.py
pawamoy: With my basic knowledge of HTML, I thought you could put anything in `data-*` attributes :slightly_smiling_face: 
I'm good with it just being a temporary data holder.

Issue #259, opened 2021-03-19
Cannot overwrite python handler or add custom handler


suned:
**Describe the bug**
I've followed the steps in the documentation to add a custom handler by placing it in `mkdocstrings/handlers/custom_handler.py`, configured it in `mkdocs.yml` and added it to a member I want to document with the handler. I intentionally didn't add a `get_handler` method to `custom_handler.py` expecting mkdocstrings to crash on `mkdocs serve`. It did not, in fact it seems to be ignoring my handler completely and proceeding to use the python handler even for members with the `handler: custom_handler` configuration set.

Moreover I added `python.py` file to `mkdocstrings/handlers/` without a `get_handler` function to test if I could overwrite the python handler. That didn't work either. As a test I modified `mkdocstrings.handlers.base.Handler` to log the path of the imported handler module, which unsurprisingly turned out to be `.venv/lib/pythonX.X/site-packages/mkdocstrings/handlers/python.py`

added minimal reproducing example to https://github.com/suned/mkdocstrings-custom-handler

**To Reproduce**
Use minimal reproducing example repo

**Expected behavior**
mkdocstrings should import my custom handler and python handler from the mkdocstrings folder in the root of my project.

**Information (please complete the following information):**
- OS: macos
- Browser: firefox
- `mkdocstrings` version: 0.15.0

**Additional context**

suned:
As a data point, it works if I set `default_handler: custom_handler` in the global config

pawamoy:
Hi @suned, thank you for the report and for your patience :sweat_smile: 

I just started looking at this, and was able to reproduce. The bug was introduced in the following commit between

https://github.com/mkdocstrings/mkdocstrings/commit/6453026fac287387090a67cce70c078377d107dd#diff-09d6e30b41d735c060b7d3f13d324a2764dcb90502715c706bdb6aaca6393081L237-L239

and

https://github.com/mkdocstrings/mkdocstrings/commit/6453026fac287387090a67cce70c078377d107dd#diff-288aa0bb4705e3681add5ce91b76e798bcf7b7ba2570563ec6e77809fa2bc166R292-R295

As you can see the `config` argument is mistakenly overwritten! 

[...]

PR #263, opened 2021-04-01
fix: Fix custom handler not being used

pawamoy:
Custom handlers configured in individual
autodoc instructions were not picked up
by mkdocstrings.

Issue #259


pawamoy:
The error seems to come from Poetry, particularly its core.

oprypin (approved):

Issue #268, opened 2021-04-11
Recursively generate documentation for entire module without creating individual markdown files


jmkd3v:
We'd like to generate documentation for our entire module without generating a markdown file for each file. I assume this is possible, but I am having trouble doing it myself.

I have tried just adding `::: package-name` to a file - which does document all contents - but it's all in one file, which is not the ideal solution. Ideally, I'd like classes and files to be automatically placed in the sidebar.

I'm using mkdocstrings version `0.15.0`. The contents of my `mkdocs.yml` file does not seem to be relevant.

oprypin:
Take a look at https://github.com/mkdocstrings/mkdocstrings/issues/179#issuecomment-784631904
(not yet sure if the issue is duplicate or not)

jmkd3v:
> Take a look at [#179 (comment)](https://github.com/mkdocstrings/mkdocstrings/issues/179#issuecomment-784631904)
> (not yet sure if the issue is duplicate or not)

While this does partially solve the problem, it's not a solution if it can't actually add the items to the nav. I assume I can write up a script that does this and then edits the nav in the mkdocs.yml file before building, but I was hoping for a more streamlined solution.

pawamoy:
So, a user could add `literate-nav` and `gen-files` to their list of plugins to automatically generate the `SUMMARY.md` and the pages for each module. Maybe we could add such snippets in the docs :slightly_smiling_face:

pawamoy:
And yes I'd consider this issue a duplicate of #179.

jmkd3v:
I'll try to write up a snippet that uses `literate-nav` to automatically generate it.

pawamoy:
See an example in #271 :stuck_out_tongue:

pawamoy:
I'll close this in favor of the original #179 issue.

PR #271, opened 2021-04-15
docs: Use literate nav to generate reference sidebar

pawamoy:
Related to #179 and #268.


oprypin:
I tried a lot of approaches, anything that would significantly simplify this. But couldn't come up with anything good.

So I can only offer a slight improvement of the code that you have here: https://github.com/mkdocstrings/mkdocstrings/commit/54b12e954baa217b29a45340c06325a37188918b

pawamoy:
Clever improvements, thanks! Any reason why you set it back to SUMMARY.md?

oprypin:
I use the name SUMMARY.md in every project. And it is a really special file so it shouldn't have a normal name.

The name originated from GitBook which uses this file name to specify the nav in the same way.

oprypin:
There's also the option of upstreaming this code.

See
https://github.com/oprypin/mkdocs-gen-files/commit/literate-nav
https://github.com/oprypin/mkdocstrings/commit/literate-nav-3

oprypin:
Based on previous mentions in chat, I figured I'd push my changes *to* this pull request. I also released the code in mkdocs-gen-files. Check it out.

pawamoy:
This is greaaaaaaaaaaaaat :smile: 
We should definitely document this in mkdocstrings docs now :stuck_out_tongue: 
But this can be done in another PR.
(I can't approve my own PR so if you want to improve, otherwise I'll force merge)

oprypin (approved):

lucmos:
@pawamoy Thanks for this PR! 
Do you have an example of how to use this feature?

It seems some files touched by this PR do not exist anymore on master

oprypin:
The relevant documentation is
https://oprypin.github.io/mkdocs-literate-nav/#programmatic-control-over-the-nav
\- and the docs are unfortunately kind of a circular reference back to this code, but oh well...

lucmos:
Thanks for the fast reply @oprypin 

It is working.
Unfortunately getting the expected result is quite hard. E.g. private modules or pages that would end up being empty must be excluded, namespaces (folders without `__init__.py`) must be taken into account, as the public API from `__all__`. I am not sure it is possible to avoid the generation of empty pages in the scope of the script...

I feel like to get it right much of the code that handles the default docs generation should be duplicated. It would be enough to somehow change the header levels in the single-page into multiple files :thinking:

pawamoy:
Hi @lucmos, thanks a lot for sharing this.

Indeed, this example works fine for simple page generation. For more complex selection/rendering options, the script can become very complex. In that case maybe the best option is to actually write the markdown pages as it allows much more granular configuration. Of course some features could be added, like respecting `__all__` :slightly_smiling_face:

Issue #102, opened 2020-04-27
Respect the order of selection.members


hSaria:
**Is your feature request related to a problem? Please describe.**
Respect the order of `selection.members` when displaying the attributes/methods of a class.

**Describe the solution you'd like**
Ordering the sections of a class according to how they were listed under `selection.members`.

**Describe alternatives you've considered**
A new option can be introduced that deals specifically with sorting, but I feel that introduces one more thing to maintain.

Another option would be to sort the methods in the same order as they were defined in the class, though I can see this becoming an issue for folks that want to sort their methods in the code in one way and their docs in another. 

**Additional context**
Assuming there is a class with the methods `add`, `subtract`, `multiply`, and `divide`, the docs currently list them in alphabetical order.

I would like to manually specify an order, maybe something like so:
```
::: module.Class
    selection:
      members:
        - add
        - subtract
        - multiply
        - divide
```

pawamoy:
I think it would be fair to respect the specified members order, though I'd like to have the opinion of others using this feature :slightly_smiling_face:  @ofek @florimondmanca @StephenBrown2 @chrieke (from #44)

Additionally, a `sorting` option would be nice. Sphinx has that: source order, alphabetical order, and group order IIRC?

ofek:
Yes please! The natural order of `members` should be the default.

StephenBrown2:
Just a note, Sphinx sorts the members by default. TIL that that's configurable!

ofek:
I'll take any way we choose 🙂

pawamoy:
Let's do this then. Just opened https://github.com/pawamoy/pytkdocs/issues/29

ofek:
@pawamoy Hmm, I see your preferred solution in https://github.com/mkdocstrings/pytkdocs/issues/29 but wouldn't it make more sense to have the renderer read the members and act accordingly? Especially since the renderer has since added a new sorting option https://github.com/mkdocstrings/mkdocstrings/pull/274

pawamoy:
See my answer here: https://github.com/mkdocstrings/pytkdocs/issues/29#issuecomment-993833625

pawamoy:
Fixed in the new handler (not the legacy one, though it's possible to fix now).

samuelcolvin:
[...]

PR #274, opened 2021-05-01
feat: Add a rendering option to change the sorting of members

joerick:
Hey! I developed this feature for my own use but figured it could be useful for others too. I wanted to make the members of a class appear as they do in the source file, instead of sorting them alphabetically. So I added this option. It works like:

```yaml
::: mkdocs_autorefs.plugin
    rendering:
      sort_members: source
```

'lexical' is the default, 'source' sets the order to that in the source file. It works by pulling the sort behaviour out of the template, and into the PythonRenderer.

This might be related to #102? Or maybe not.

### DIFF
--- a/docs/reference/autorefs/plugin.md
+++ b/docs/reference/autorefs/plugin.md
@@ -1,3 +1,1 @@
 ::: mkdocs_autorefs.plugin
-    rendering:
-      members_order: source
--- a/docs/reference/autorefs/plugin.md
+++ b/docs/reference/autorefs/plugin.md
@@ -1,1 +1,3 @@
 ::: mkdocs_autorefs.plugin
+    rendering:
+      sort_members: source
--- a/docs/reference/autorefs/plugin.md
+++ b/docs/reference/autorefs/plugin.md
@@ -1,3 +1,3 @@
 ::: mkdocs_autorefs.plugin
     rendering:
-      sort_members: source
+      members_order: source
--- a/src/mkdocstrings/handlers/python.py
+++ b/src/mkdocstrings/handlers/python.py
@@ -49,6 +49,7 @@ class PythonRenderer(BaseRenderer):
         "show_bases": True,
         "group_by_category": True,
         "heading_level": 2,
+        "sort_members": "lexical",
     }
     """The default rendering options.
 
@@ -66,6 +67,7 @@ class PythonRenderer(BaseRenderer):
     **`show_bases`** | `bool` | Show the base classes of a class. | `True`
     **`group_by_category`** | `bool` | Group the object's children by categories: attributes, classes, functions, methods, and modules. | `True`
     **`heading_level`** | `int` | The initial heading level to use. | `2`
+    **`sort_members`** | `str` | The sort option to use. Options: `lexical` - order by the member name, `source` - order members as they appear in the source file | `lexical`
     """  # noqa: E501
 
     def render(self, data: CollectorItem, config: dict) -> str:  # noqa: D102 (ignore missing docstring)
@@ -77,6 +79,8 @@ class PythonRenderer(BaseRenderer):
         # of the rendering recursion. Therefore, it's easier to use it as a plain value
         # than as an item in a dictionary.
         heading_level = final_config["heading_level"]
+
+        sort_object(data, sort_style=final_config["sort_members"])
 
         return template.render(
             **{"config": final_config, data["category"]: data, "heading_level": heading_level, "root": True},
@@ -324,3 +328,30 @@ def rebuild_category_lists(obj: dict) -> None:
     obj["children"] = [child for _, child in obj["children"].items()]
     for child in obj["children"]:
         rebuild_category_lists(child)
+
+
+def sort_object(obj: CollectorItem, sort_style: str) -> None:
+    """Sort the collected object's children.
+
+    Sorts the object's children list, then each category separately, and then recurses into each.
+
+    Arguments:
+        obj: The collected object, as a dict. Note that this argument is mutated.
+        sort_style: How to sort the children lists - 'lexical' or 'source'.
+    """
+
+    def sort_key(a: dict) -> Any:
+        if sort_style == "lexical":
+            return a.get("name")
+        elif sort_style == "source":
+            return a.get("source", {}).get("line_start", 0)
+        else:
+            raise Exception("unknown sort_style")
+
+    obj["children"].sort(key=sort_key)
+
+    for category in ("attributes", "classes", "functions", "methods", "modules"):
+        obj[category].sort(key=sort_key)
+
+    for child in obj["children"]:
+        sort_object(child, sort_style=sort_style)
--- a/src/mkdocstrings/handlers/python.py
+++ b/src/mkdocstrings/handlers/python.py
@@ -18,6 +18,7 @@ from markupsafe import Markup
 from mkdocstrings.handlers.base import BaseCollector, BaseHandler, BaseRenderer, CollectionError, CollectorItem
 from mkdocstrings.inventory import Inventory
 from mkdocstrings.loggers import get_logger
+from mkdocstrings.extension import PluginError
 
 log = get_logger(__name__)
 
@@ -346,7 +347,7 @@ def sort_object(obj: CollectorItem, sort_style: str) -> None:
         elif sort_style == "source":
             return a.get("source", {}).get("line_start", 0)
         else:
-            raise Exception("unknown sort_style")
+            raise PluginError("unknown sort_style")
 
     obj["children"].sort(key=sort_key)
 
--- a/src/mkdocstrings/handlers/python.py
+++ b/src/mkdocstrings/handlers/python.py
@@ -49,7 +49,7 @@ class PythonRenderer(BaseRenderer):
         "show_bases": True,
         "group_by_category": True,
         "heading_level": 2,
-        "sort_members": "lexical",
+        "members_order": "alphabetical",
     }
     """The default rendering options.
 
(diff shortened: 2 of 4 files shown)


src/mkdocstrings/handlers/python.py
```diff
@@ -45,6 +45,7 @@ class PythonRenderer(BaseRenderer):
         "show_source": True,
         "group_by_category": True,
         "heading_level": 2,
+        "sort_members": "lexical",
```
oprypin, line 48: I'd prefer the more general word "order" rather than "sort". I think "keeping original order" is no longer covered by the word "sort".

Also for "lexical", maybe you mean "lexicographic". But isn't "alphabetic" a much simpler term for this?

So maybe this would look something like:
```yaml
members_order: source
members_order: alphabetic
```
(but I'm not sure even about that specific naming, `members_order` looks a bit weird)

src/mkdocstrings/handlers/python.py
```diff
@@ -290,3 +294,30 @@ def rebuild_category_lists(obj: dict) -> None:
...
+        else:
+            raise Exception("unknown sort_style")
+
+    obj["children"].sort(key=sort_key)
+
+    for category in ("attributes", "classes", "functions", "methods", "modules"):
+        obj[category].sort(key=sort_key)
+
+    for child in obj["children"]:
+        sort_object(child, sort_style=sort_style)
```
oprypin, line 323: There's probably some more specific way to throw errors. Maybe `PluginError`?

In the future we could consider validating configs at the very beginning, but I think the infrastructure is currently not there, so disregard that part.

src/mkdocstrings/handlers/python.py
```diff
@@ -290,3 +294,30 @@ def rebuild_category_lists(obj: dict) -> None:
...
+    Sorts the object's children list, then each category separately, and then recurses into each.
+
+    Arguments:
+        obj: The collected object, as a dict. Note that this argument is mutated.
+        sort_style: How to sort the children lists - 'lexical' or 'source'.
+    """
+
+    def sort_key(a: dict) -> Any:
+        if sort_style == "lexical":
+            return a.get("name")
```
oprypin, line 311: This can break if something is actually missing a name. Then inside `sort` you'll get

```
TypeError: '<' not supported between instances of 'NoneType' and 'str'
```

Maybe this case needs to be handled better, or maybe just use `a["name"]` to get the more explicit error ... or could it be `a.get("name", "")` analogously to your next line?

Although I'd probably prefer non-sortable objects to appear at the end rather than at the beginning.

src/mkdocstrings/handlers/python.py
```diff
@@ -290,3 +294,30 @@ def rebuild_category_lists(obj: dict) -> None:
     obj["children"] = [child for _, child in obj["children"].items()]
     for child in obj["children"]:
         rebuild_category_lists(child)
+
+
+def sort_object(obj: CollectorItem, sort_style: str) -> None:
```
oprypin, line 299: I'll have to point out that it's possible to write unittests for this :D

oprypin (commented):
Thanks for your contribution! It makes sense and I don't see a problem with this.

I will let @pawamoy lead the review, though.

I am just leaving some comments on the details, but no need to act on them immediately.

src/mkdocstrings/handlers/python.py
pawamoy: Sphinx's autodoc calls it [`autodoc_member_order`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_member_order) so I guess `members_order` is fine. As for the values, I'd vote for `alphabetical` and `source` :slightly_smiling_face:

src/mkdocstrings/handlers/python.py
```diff
@@ -290,3 +294,30 @@ def rebuild_category_lists(obj: dict) -> None:
...
+        sort_style: How to sort the children lists - 'lexical' or 'source'.
+    """
+
+    def sort_key(a: dict) -> Any:
+        if sort_style == "lexical":
+            return a.get("name")
+        elif sort_style == "source":
+            return a.get("source", {}).get("line_start", 0)
+        else:
+            raise Exception("unknown sort_style")
```
pawamoy, line 315: Rather than defining a function containing the condition, I think it would be preferable to run the condition once to define the proper sorting function. For example:

```python
import operator

if sort_style == "alphabetical":
    sort_function = operator.itemgetter("name")
elif sort_style == "source":
    sort_function = lambda item: item.get("source", {}).get("line_start", 0)
```

Maybe we could even write those once and for all as private functions, and pass one or the other directly to the `sort_object` function based on the config.

pawamoy (commented):
Hi @joerick, thank you for this PR :slightly_smiling_face: 
I think a lot of people will love that feature!

Just added a few comments. I'll try to be more thorough later.

src/mkdocstrings/handlers/python.py
joerick: `PluginError` makes sense to me!

src/mkdocstrings/handlers/python.py
joerick: I have a solution to this in 99e7dc4, curious what you think!

src/mkdocstrings/handlers/python.py
joerick: Yes, I wasn't sure how to set that up. Is there a similar test (e.g. for a rendering option) I could use as a template?

joerick:
Thanks for the reviews! I've made a few edits based on your feedback.

docs/reference/autorefs/plugin.md
```diff
@@ -1 +1,3 @@
 ::: mkdocs_autorefs.plugin
+    rendering:
+      sort_members: source
```
pawamoy, line 3:
```suggestion
      members_order: source
```

src/mkdocstrings/handlers/python.py
```diff
@@ -61,6 +63,7 @@ class PythonRenderer(BaseRenderer):
     **`show_source`** | `bool` | Show the source code of this object. | `True`
     **`group_by_category`** | `bool` | Group the object's children by categories: attributes, classes, functions, methods, and modules. | `True`
     **`heading_level`** | `int` | The initial heading level to use. | `2`
+    **`members_order`** | `str` | The sort option to use. Options: `alphabetical` - order by the member name, `source` - order members as they appear in the source file | `alphabetical`
```
pawamoy, line 66:
```suggestion
    **`members_order`** | `str` | The members ordering to use. Options: `alphabetical` - order by the members names, `source` - order members as they appear in the source file. | `alphabetical`
```

src/mkdocstrings/handlers/python.py
```diff
@@ -290,3 +295,33 @@ def rebuild_category_lists(obj: dict) -> None:
...
+    if sort_style == "alphabetical":
+        # sort_last is a string that contains the final unicode character, so
+        # if 'name' isn't found on the object, the item will go to the end of
+        # the list.
+        sort_last = chr(sys.maxunicode)
+        sort_function = lambda item: item.get("name", sort_last)
+    elif sort_style == "source":
+        sort_function = lambda item: item.get("source", {}).get("line_start", 0)
+    else:
+        raise PluginError('unknown sort_style')
```
pawamoy, line 319: @oprypin @joerick do you think we should optimize this again by refactoring the two lambdas into proper (private) functions? One function or the other would then be passed directly to `sort_object` instead of re-computing it in each step of the recursion. I know it might be premature optimization, but it's easy to do, so why not.

This would also allow getting the PluginError out of this function.

src/mkdocstrings/handlers/python.py
```diff
@@ -290,3 +298,35 @@ def rebuild_category_lists(obj: dict) -> None:
...
+
+    for category in ("attributes", "classes", "functions", "methods", "modules"):
+        obj[category].sort(key=sort_function)
+
+    for child in obj["children"]:
+        sort_object(child, sort_function=sort_function)
+
+
+def _sort_key_alphabetical(item: CollectorItem) -> Any:
+    """Returns a sort key for 'alphabetical' sorting of CollectorItems"""
```
pawamoy, line 323:
```suggestion
    """Return a sort key for 'alphabetical' sorting of CollectorItems."""
```

src/mkdocstrings/handlers/python.py
```diff
@@ -290,3 +298,35 @@ def rebuild_category_lists(obj: dict) -> None:
...
+    # chr(sys.maxunicode) is a string that contains the final unicode
+    # character, so if 'name' isn't found on the object, the item will go to
+    # the end of the list.
+    return item.get("name", chr(sys.maxunicode))
+
+
+def _sort_key_source(item: CollectorItem) -> Any:
+    """Returns a sort key for 'source' sorting of CollectorItems"""
```
pawamoy, line 331:
```suggestion
    """Return a sort key for 'source' sorting of CollectorItems."""
```

src/mkdocstrings/handlers/python.py
```diff
@@ -72,6 +75,16 @@ def render(self, data: CollectorItem, config: dict) -> str:  # noqa: D102 (ignor
...
         # than as an item in a dictionary.
         heading_level = final_config["heading_level"]
+        members_order = final_config["members_order"]
+
+        if members_order == "alphabetical":
+            sort_function = _sort_key_alphabetical
+        elif members_order == "source":
+            sort_function = _sort_key_source
+        else:
+            raise PluginError("unknown members_order")
```
pawamoy, line 85:
```suggestion
            raise PluginError(f"Unknown members_order '{members_order}', choose between 'alphabetical' and 'source'.")
```

pawamoy (changes requested):
Thank you for the update! A few suggestions :slightly_smiling_face: 
Please add a test for each sorting function as well, and we should be good to go!
Let me know if you need any advice about writing the tests.

pawamoy (approved):

camold:
When will this appear in a public version on pypi? The last version (0.15) dates from February, unfortunately.

pawamoy:
We should be able to release 0.16 soon. Just polishing docs and we're good to go.

camold:
Use
members_order: source
instead. The option has been renamed - see 
https://github.com/mkdocstrings/mkdocstrings/blob/master/src/mkdocstrings/handlers/python.py#L53

ofek:
@joerick Trying this now, just to confirm: was your intention to order by appearance rather than line number of definition?

joerick:
hey @ofek! The intention was to render docs in the order that the classes/methods appear in the .py file

PR #275, opened 2021-05-02
refactor: Move writing extra files to an earlier stage in the build

oprypin:


oprypin:
This step does not depend on anything that happens the HTML-rendering steps, only on the Markdown-rendering steps, so it's fine to move earlier.
    
An example of why it's bad that the files do not exist during the HTML-rendering steps can be seen in a hypothetical commit https://github.com/oprypin/mkdocs/commit/a677dd7f591d3f6587cde0409a2b2d43cb889ade - basing the HTML content on the content of those files fails.

And this is, of course, inspired by my diagram https://github.com/mkdocs/mkdocs/pull/2387

pawamoy (approved):
Fine by me!

PR #277, opened 2021-05-02
feat: Support loading external inventories and linking to them

oprypin:
Example of how an external handler can be implemented:

* https://github.com/mkdocstrings/crystal/compare/inv#commit_comments_bucket

* ```yaml
  plugins:
  - mkdocstrings:
      handlers:
        crystal:
          import:
            - https://crystal-lang.org/api/latest/index.json
  ```

Alternative to https://github.com/mkdocstrings/autorefs/pull/4

### DIFF
--- a/src/mkdocstrings/plugin.py
+++ b/src/mkdocstrings/plugin.py
@@ -200,7 +200,7 @@ class MkdocstringsPlugin(BasePlugin):
             inv_loader = concurrent.futures.ThreadPoolExecutor(4)
             for handler_name, import_item in to_import:
                 future = inv_loader.submit(
-                    self._load_inventory, self.get_handler(handler_name).load_inventory, import_item
+                    self._load_inventory, self.get_handler(handler_name).load_inventory, **import_item
                 )
                 self._inv_futures.append(future)
             inv_loader.shutdown(wait=False)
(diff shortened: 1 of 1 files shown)


pawamoy (commented):
We definitely need to be able to specify the base URL for (Python) inventories, since these inventories could contain relative links only, and the base URL could not always be inferred from the inventory URL.

Other than that: very nice code :slightly_smiling_face:

oprypin:
Ah that can actually work even in this state.

https://github.com/mkdocstrings/mkdocstrings/pull/277/files#diff-3968cbe3b57ebe74c208010a8eb7087459b75fe8d4e7bc5a29a30a882c5e4281R165-R167

Even though it says that `to_import` items are "URLs", actually it can be a freeform argument, and each handler is free to deal with it.

So for Python handler maybe it will support

```yaml
        import:
          - https://example.org/objects.inv
```
```yaml
        import:
          - [https://example.org/objects.inv, https://example.org/]
```

oprypin:
Wait no, what I said is not true, the URL *is* used explicitly :facepalm:

oprypin:
Now I pushed code to actually allow this.

> So for Python handler maybe it will support

```yaml
        import:
          - https://example.org/objects.inv
```
```yaml
        import:
          - url: https://example.org/objects.inv
```
The above two are equivalent, and the 3rd adds another arg. Handling named args will be up to each handler.
```yaml
        import:
          - url: https://example.org/objects.inv
            base_url: https://example.org/
```

src/mkdocstrings/plugin.py
```diff
@@ -211,6 +234,12 @@ def on_env(self, env, config: Config, **kwargs):
                 inv_contents = self.handlers.inventory.format_sphinx()
                 write_file(inv_contents, os.path.join(config["site_dir"], "objects.inv"))
 
+        if self._inv_futures:
+            concurrent.futures.wait(self._inv_futures, timeout=30)
```
pawamoy, line 238: Should we log something just before waiting in case the downloads are slow?

pawamoy (approved):
OK that looks great! Do you want to add tests or should we merge? I guess actual inventory loading could be tested once the Python handler knows how to do it.

oprypin:
I'm quite excited to merge, let's add tests a bit later

PR #283, opened 2021-05-08
build: Looks like we're ready for the upcoming major Jinja release

oprypin:
and same for MarkupSafe.

They mainly just drop Python 2.
* https://jinja.palletsprojects.com/en/master/changes/#version-3-0-0
* https://markupsafe.palletsprojects.com/en/master/changes/#version-2-0-0

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ packages = [ { include = "mkdocstrings", from = "src" } ]
 
 [tool.poetry.dependencies]
 python = "^3.6"
-Jinja2 = "^2.11.1, <4.0"
+Jinja2 = ">=2.11.1, <4.0"
 Markdown = "^3.3"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -16,9 +16,9 @@ packages = [ { include = "mkdocstrings", from = "src" } ]
 
 [tool.poetry.dependencies]
 python = "^3.6"
-Jinja2 = "^2.11"
+Jinja2 = "^2.11.1, <4.0"
```
pawamoy, line 19:
```suggestion
Jinja2 = ">=2.11.1, <4.0"
```

oprypin: Wow good catch

pawamoy:
Thanks for keeping us up-to-date :slightly_smiling_face:

Issue #269, opened 2021-04-13
Show base class of python classes


valmar:
I am testing mkdocs for the documentation of my python project and I am up until now impressed by its combination with mkdocstrings.

I was wondering if would be possible please to have the base class for derived classes displayed in the auto-generated documentation. In other words, if we have a class like:

```
class DerivedClass(BaseClass):
  pass
```
I would like to have somewhere in the documentation of DerivedClass something like: "Base class: BaseClass". I am open to any option or workaround to display this information: modifying the templates, etc.

Thank you in advance

pawamoy:
Hello, glad to hear you enjoy this project :heart: 

It would be a nice addition indeed. This could be relatively easy to achieve in pytkdocs, by returning the parent classes (from `__mro__`) in the serialized output. They could then be rendered by mkdocstrings. We would add an option which defaults to false.

PR #297, opened 2021-06-19
Optionally show base classes

bkoropoff:
I'm not sure if this is the best way to present them, or if it should be on by default.  Feedback welcome.

Fixes https://github.com/mkdocstrings/mkdocstrings/issues/269
Requires https://github.com/mkdocstrings/pytkdocs/pull/108
Best with https://github.com/mkdocstrings/autorefs/pull/10 so that you can hover over a base class to see the fully qualified name

Screenshot:

shot

### DIFF
--- a/src/mkdocstrings/handlers/python.py
+++ b/src/mkdocstrings/handlers/python.py
@@ -13,6 +13,7 @@ from subprocess import PIPE, Popen  # noqa: S404 (what other option, more secure
 from typing import Any, BinaryIO, Iterator, List, Optional, Tuple
 
 from markdown import Markdown
+from markupsafe import Markup
 
 from mkdocstrings.handlers.base import BaseCollector, BaseHandler, BaseRenderer, CollectionError, CollectorItem
 from mkdocstrings.inventory import Inventory
@@ -45,6 +46,7 @@ class PythonRenderer(BaseRenderer):
         "show_if_no_docstring": False,
         "show_signature_annotations": False,
         "show_source": True,
+        "show_bases": True,
         "group_by_category": True,
         "heading_level": 2,
     }
@@ -61,6 +63,7 @@ class PythonRenderer(BaseRenderer):
     **`show_if_no_docstring`** | `bool` | Show the object heading even if it has no docstring or children with docstrings. | `False`
     **`show_signature_annotations`** | `bool` | Show the type annotations in methods and functions signatures. | `False`
     **`show_source`** | `bool` | Show the source code of this object. | `True`
+    **`show_bases`** | `bool` | Show the base classes of a class. | `True`
     **`group_by_category`** | `bool` | Group the object's children by categories: attributes, classes, functions, methods, and modules. | `True`
     **`heading_level`** | `int` | The initial heading level to use. | `2`
     """  # noqa: E501
@@ -87,6 +90,12 @@ class PythonRenderer(BaseRenderer):
         self.env.trim_blocks = True
         self.env.lstrip_blocks = True
         self.env.keep_trailing_newline = False
+        self.env.filters["brief_xref"] = self.do_brief_xref
+
+    def do_brief_xref(self, path: str) -> Markup:
+        """Filter to create cross-reference with brief text and full identifier as hover text."""
+        brief = path.split(".")[-1]
+        return Markup("<span data-autorefs-optional-hover={path}>{brief}</span>").format(path=path, brief=brief)
 
 
 class PythonCollector(BaseCollector):
--- a/src/mkdocstrings/templates/python/material/class.html
+++ b/src/mkdocstrings/templates/python/material/class.html
@@ -22,7 +22,14 @@
           class="doc doc-heading",
           toc_label=class.name) %}
 
-        <code>{% if show_full_path %}{{ class.path }}{% else %}{{ class.name }}{% endif %}</code>
+        <code>
+          {% if show_full_path %}{{ class.path }}{% else %}{{ class.name }}{% endif %}
+          {% if config.show_bases and class.bases != ['object'] %}
+            ({% for base in class.bases -%}
+              {{ base|brief_xref() }}{% if not loop.last %}, {% endif %}
+             {% endfor %})
+          {% endif %}
+        </code>
 
         {% with properties = class.properties %}
           {% include "properties.html" with context %}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ Jinja2 = ">=2.11.1, <4.0"
 Markdown = "^3.3"
 MarkupSafe = ">=1.1, <3.0"
 mkdocs = "^1.1.1"
-mkdocs-autorefs = ">=0.1, <0.3"
+mkdocs-autorefs = ">=0.1, <0.4"
 pymdown-extensions = ">=6.3, <9.0"
 pytkdocs = ">=0.2.0, <0.12.0"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ MarkupSafe = ">=1.1, <3.0"
 mkdocs = "^1.1.1"
 mkdocs-autorefs = ">=0.1, <0.4"
 pymdown-extensions = ">=6.3, <9.0"
-pytkdocs = ">=0.2.0, <0.12.0"
+pytkdocs = ">=0.2.0, <0.13.0"
 
 [tool.poetry.dev-dependencies]
 # formatting, quality, tests


bkoropoff:
@pawamoy

pawamoy:
Hey @bkoropoff, thanks a lot for your contributions, I'll try to review this in the next few days.

src/mkdocstrings/handlers/python.py
```diff
@@ -45,6 +46,7 @@ class PythonRenderer(BaseRenderer):
         "show_if_no_docstring": False,
         "show_signature_annotations": False,
         "show_source": True,
+        "show_bases": True,
```
pawamoy, line 49: I was going to ask to default to false as to not alter previous output, but I guess showing base classes is the sane default option, since it doesn't break anything and is nice to see. If you have an opinion on this please share!

src/mkdocstrings/handlers/python.py
```diff
@@ -87,7 +90,11 @@ def update_env(self, md: Markdown, config: dict) -> None:  # noqa: D102 (ignore
         self.env.trim_blocks = True
         self.env.lstrip_blocks = True
         self.env.keep_trailing_newline = False
+        self.env.filters["brief_xref"] = self.do_brief_xref
 
+    def do_brief_xref(self, path: str) -> Markup:
+        brief = path.split(".")[-1]
+        return Markup(f"<span data-autorefs-optional=\"{path}\">{brief}</span>")
```
pawamoy, line 97: Nice way to handle this, but the user has no control over this. I'm willing to merge anyway because I consider it a first step to allow users to shorten paths like that. We can still add some special markup later that will take advantage of this filter, so users can control it.

src/mkdocstrings/templates/python/material/class.html
```diff
@@ -22,7 +22,14 @@
           class="doc doc-heading",
           toc_label=class.name) %}
 
-        <code>{% if show_full_path %}{{ class.path }}{% else %}{{ class.name }}{% endif %}</code>
+        <code>
+          {% if show_full_path %}{{ class.path }}{% else %}{{ class.name }}{% endif %}
+          {% if config.show_bases %}
+            ({% for base in class.bases -%}
+              {{ base|brief_xref() }}{% if base != class.bases[-1] %}, {% endif %}
```
pawamoy, line 29: I believe the intent here is to skip for the comma after the last item? Then you can use `{% if not loop.last %}` instead. See https://jinja2docs.readthedocs.io/en/stable/templates.html#for

pawamoy (changes requested):
Two comments and one thing you can update.
Also make sure to run `make format` and `make check` locally and fix any issues. I see this in CI:

```
  src/mkdocstrings/handlers/python.py:95:1: D102 Missing docstring in public method
  src/mkdocstrings/handlers/python.py:97:24: BLK100 Black would make changes.
  src/mkdocstrings/handlers/python.py:99:1: E302 expected 2 blank lines, found 1
```

Thanks a lot :slightly_smiling_face:

src/mkdocstrings/templates/python/material/class.html
bkoropoff: Yeah, thanks.  I hadn't used jinja before this.

src/mkdocstrings/handlers/python.py
bkoropoff: Yeah, it should probably be an option whether to shorten implicit cross-references (or even create them at all).  This seems like a good default, though, especially with the autorefs change to show the full path on hover.

src/mkdocstrings/handlers/python.py
bkoropoff: You want to know what the base classes are more often than not, since it tells where you can use it (anything that takes a `BaseRenderer` can take a `PythonRenderer`, since it's an interface), and what additional methods are available.  You could turn on `show_inherited_members`, but this brings in a bunch of noise from builtin Python classes.

Sometimes a base class is an implementation detail, but that's rarer.

src/mkdocstrings/templates/python/material/class.html
```diff
@@ -22,7 +22,14 @@
           class="doc doc-heading",
           toc_label=class.name) %}
 
-        <code>{% if show_full_path %}{{ class.path }}{% else %}{{ class.name }}{% endif %}</code>
+        <code>
+          {% if show_full_path %}{{ class.path }}{% else %}{{ class.name }}{% endif %}
+          {% if config.show_bases %}
+            ({% for base in class.bases -%}
```
pawamoy, line 28: One more thing: do we want to render `()` when there aren't any base class? I would say no. In that case an additional condition on `class.bases`  would be used before the loop.

pawamoy: Maybe you can combine both conditions like so:

```
{% if config.show_bases and class.bases %}
```

bkoropoff: A Python 3 class with no explicit bases implicitly gets `object` as a base, so this can't happen.  We could omit the list entirely if `object` is the only base.

pawamoy: Ah, true, my bad. Then yeah, we must decide if showing `(object)` has value or not. I'd say it doesn't, what do you think?

bkoropoff:
Updated.  https://github.com/mkdocstrings/autorefs/pull/10 is now a hard requirement because it uses the new `autorefs-optional-hover` reference type.  Base classes are omitted if the only one is `object`.

src/mkdocstrings/templates/python/material/class.html
bkoropoff: The official python 3 docs seem to omit it in code and use the implicit behavior, so I changed it to match.

src/mkdocstrings/handlers/python.py
bkoropoff: I'll leave it on by default unless you have strong objections.

oprypin (commented):
Great to see such an advanced contribution :)

src/mkdocstrings/handlers/python.py
```diff
@@ -87,6 +90,12 @@ def update_env(self, md: Markdown, config: dict) -> None:  # noqa: D102 (ignore
         self.env.trim_blocks = True
         self.env.lstrip_blocks = True
         self.env.keep_trailing_newline = False
+        self.env.filters["brief_xref"] = self.do_brief_xref
+
+    def do_brief_xref(self, path: str) -> Markup:
+        """Filter to create cross-reference with brief text and full identifier as hover text."""
+        brief = path.split(".")[-1]
+        return Markup(f'<span data-autorefs-optional-hover="{path}">{brief}</span>')
```
oprypin, line 98: Both of these values need to be escaped, and I believe this is one way to achieve that:
```suggestion
        return Markup('<span data-autorefs-optional-hover="{}">{}</span>').format(path, brief)
```

oprypin:
On an organizational note: a release of autorefs will need to be made, and the dependency will need to be bumped here

src/mkdocstrings/handlers/python.py
pawamoy: Do you mean we need to escape potential double-quotes in path and brief for example?
I think both format and the f-string give the exact same result :confused:

oprypin:
```pycon
>>> from markupsafe import Markup
>>> path = 'foo&bar'
>>> brief = '"bar"'
>>> Markup(f'<span data-autorefs-optional-hover="{path}">{brief}</span>')
Markup('<span data-autorefs-optional-hover="foo&bar">"bar"</span>')
>>> Markup('<span data-autorefs-optional-hover="{}">{}</span>').format(path, brief)
Markup('<span data-autorefs-optional-hover="foo&amp;bar">&#34;bar&#34;</span>')
```

pawamoy: Aaaaah my bad, I thought it was about the `format` *builtin*, but it's rather the `format` method of Markup. Thanks for the example.

bkoropoff:
@pawamoy @oprypin Should be resolved now

bkoropoff:
@pawamoy @oprypin Is this ready to merge?

pawamoy:
@bkoropoff sorry for the delay, I'll get to this tomorrow.

pawamoy:
> On an organizational note: a release of autorefs will need to be made, and the dependency will need to be bumped here

Just released mkdocs-autorefs 0.3.0 with @bkoropoff's changes. @bkoropoff can you increase the version range in pyproject.toml to accept version 0.3.0?

bkoropoff:
Yeah.  pytkdocs also needs to be released/bumped, right?

pawamoy:
Yes indeed, I forgot that you sent a PR to pytkdocs too!

pawamoy (dismissed):

pawamoy (approved):

PR #310, opened 2021-08-22
Document loading and generating inventories

pawamoy:
Not the best docs I've written, but it's a start.


astrojuanlu:
These docs aren't live yet, right?

PR #327, opened 2021-10-03
Support `pymdown-extensions` v9.x

ofek:
https://facelessuser.github.io/pymdown-extensions/about/changelog/#90


ofek:
Just FYI until this is released, the latest versions of `mkdocs-material` cannot be used: https://github.com/squidfunk/mkdocs-material/blob/7.3.1/requirements.txt#L25

pawamoy:
Thank you @ofek!

Issue #205, opened 2020-12-30
Support for documentation of **kwargs


sansbacon:
I'm using Google-style docstrings and providing documentation on **kwargs
```
       Crossover operation to generate new population

        Args:
            None
            
        Keyword Args:
            agg (bool): if True, then aggregate multiple crossovers, otherwise is sequential.
            population (np.ndarray): the population to cross over, is 2D array
            population_fitness (np.ndarray): the fitness of the population to crossover, is 1D array
            pctl (int): the percentile fitness (1-99) to use in crossover, default 50 

        Returns:
            np.ndarray

```

However, the docstring ends up looking like this when processed by mkdocstrings

```
crossover(self, **kwargs)
Crossover operation to generate new population

Keyword Args: agg (bool): if True, then aggregate multiple crossovers, otherwise is sequential. population (np.ndarray): the population to cross over, is 2D array population_fitness (np.ndarray): the fitness of the population to crossover, is 1D array pctl (int): the percentile fitness (1-99) to use in crossover, default 50

!!! returns np.ndarray
```

Is there a workaround or is it possible to support the Keyword Arguments section?

Best regards,
Eric

pawamoy:
I don't see any reference to a possible "Keyword Args" section in the [Google style guide](https://google.github.io/styleguide/pyguide.html) for Python. If you do, could you share a link?

In any case, you can document your `**kwargs` in the "Args" section, like any other argument:

```
       Crossover operation to generate new population

        Args:
            **kwargs:
                - agg (bool): if True, then aggregate multiple crossovers, otherwise is sequential.
                - population (np.ndarray): the population to cross over, is 2D array
                - population_fitness (np.ndarray): the fitness of the population to crossover, is 1D array
                - pctl (int): the percentile fitness (1-99) to use in crossover, default 50 

        Returns:
            np.ndarray
```

Note however that this is written in Markdown, and won't appear as *a table in a table* (unless you use the markup for a table).

gegnew:
It's supported by [napoleon](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/index.html#docstring-sections), and I'd definitely second the option to specify a Keyword Args section.

As it is, if you have a lot of kwargs that you don't want in your function definition, but you do want in the docstring...

```
def my_function(arg1, arg2, **kwargs):
    """
    Args:
        arg1: foo
        arg2: bar
        kwarg1 (str) = 'one'
        kwarg2 (int) = 2
    """
[...]

PR #331, opened 2021-10-08
Support keyword args and yields sections

### LINKED ISSUE #324: Support "Yields" in docstrings
Per the [Google styleguide](https://google.github.io/styleguide/pyguide.html#doc-function-returns), the keyword `Yields` is used for generators. However, when using `Yields` in a docstring, it is not parsed by `mkdocstrings`:

image

I'd like `Yields` to be parsed in the same manner that `Returns` is. :)

pawamoy:
Closes #324 and #205

### DIFF
--- a/src/mkdocstrings/templates/python/material/docstring.html
+++ b/src/mkdocstrings/templates/python/material/docstring.html
@@ -11,9 +11,17 @@
       {% with parameters = section.value %}
         {% include "parameters.html" with context %}
       {% endwith %}
+    {% elif section.type == "keyword_args" %}
+      {% with kwargs = section.value %}
+        {% include "keyword_args.html" with context %}
+      {% endwith %}
     {% elif section.type == "exceptions" %}
       {% with exceptions = section.value %}
         {% include "exceptions.html" with context %}
+      {% endwith %}
+    {% elif section.type == "yield" %}
+      {% with yield = section.value %}
+        {% include "yield.html" with context %}
       {% endwith %}
     {% elif section.type == "return" %}
       {% with return = section.value %}
--- a/src/mkdocstrings/templates/python/material/keyword_args.html
+++ b/src/mkdocstrings/templates/python/material/keyword_args.html
@@ -1,0 +1,20 @@
+{{ log.debug() }}
+<p><strong>Keyword arguments:</strong></p>
+<table>
+  <thead>
+    <tr>
+      <th>Name</th>
+      <th>Type</th>
+      <th>Description</th>
+    </tr>
+  </thead>
+  <tbody>
+    {% for kwarg in kwargs %}
+      <tr>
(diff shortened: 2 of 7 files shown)


pawamoy:
CI needs #330 to pass.

oprypin:
Actually... I don't understand the purpose behind a "keyword args" section. The descriptions of all parameters can be within one section, why should they be separated?
The actual viability of passing arguments should be visible from the function's signature.

pawamoy:
The purpose of the Keyword Argument section is to be able to document arguments that can be passed through `**kwargs`, and to enjoy the same formatting as for regular Arguments sections.

Consider this:

```python
def function(a, b, **kwargs):
    c = kwargs.pop("c")
    d = kwargs.pop("d")
    # use a, b, c, d
```

You can document `a`, `b`, and `**kwargs` in the Arguments section, but not `c` and `d` directly, as they are not part of the signature (linters not happy). You could document them in the `**kwargs` description, but you'd have to use plain Markdown formatting for anything a bit elaborate. Instead, with a Keyword Arguments section, you can document `c` and `d` just like any other argument:

```
Keyword Arguments:
    c (int): Description.
    d (list[str] | None): Description
```

oprypin:
Wait, I thought that Python has eliminated all cases where you'd have to rely on `**kwargs` to serve particularly-named keyword-only arguments. Wouldn't it be good to just hope that the last usages of this will die out soon and encourage writing the arguments directly?
```python
def function(a, b, *, c, d):
```

pawamoy:
You have good arguments as always @oprypin :slightly_smiling_face: 
I don't know if it's reasonable to expect that this use of `**kwargs` will disappear (its only use apart from swallowing unwanted arguments?). Support for this was explicitly requested by users :confused:

oprypin:
Well, if supporting this never pops up in any way except when people write `Keyword arguments:`, that's fine. (I was afraid that this would *automatically* create such sections where it didn't before). But I wouldn't change correct wording to wrong just to match this rare use case.

BTW they could instead describe `kwargs` as its own item within `Args`, with normal Markdown sub-lists in it

pawamoy:
> BTW they could instead describe kwargs as its own item within Args, with normal Markdown sub-lists in it

Yes that's what I initially replied in the original issue, #205.

> I was afraid that this would automatically create such sections where it didn't before

Oh no, it's only rendered when a "Keyword arguments" section is found indeed.

> But I wouldn't change correct wording to wrong just to match this rare use case.

Agreed. Then I'll totally revise what I said earlier, and change the section title to "Keyword parameters". Maybe "Other parameters", from Numpydoc, would be even more appropriate.

oprypin:
OK! Now no issue from my side.
I will *not* do a super careful review of all the content, so that's all.
Thanks

pawamoy:
Failures came from a precedent version of PDM which fixed the KeyError issue.

PR #350, opened 2021-11-28
refactor: Return multiple identifiers from fallback method

pawamoy:
Since an object can have aliases (different identifiers
leading to it), and since users sometimes want to
render an object using one of its aliases instead of its
canonical identifier, we make sure to register every
identifier associated to an object so that autorefs
can find it when fixing cross-references.

Issue mkdocstrings/autorefs#11
Needs PR https://github.com/mkdocstrings/autorefs/pull/12

### DIFF
--- a/src/mkdocstrings/handlers/base.py
+++ b/src/mkdocstrings/handlers/base.py
@@ -125,18 +125,20 @@ class BaseRenderer(ABC):
             The rendered template as HTML.
         """  # noqa: DAR202 (excess return section)
 
-    def get_anchor(self, data: CollectorItem) -> Optional[str]:
-        """Return the canonical identifier (HTML anchor) for a collected item.
-
-        This must match what the renderer would've actually rendered,
-        e.g. if rendering the item contains `<h2 id="foo">...` then the return value should be "foo".
+    def get_anchors(self, data: CollectorItem) -> Sequence[str]:
+        """Return the possible identifiers (HTML anchors) for a collected item.
 
         Arguments:
             data: The collected data.
 
         Returns:
-            The HTML anchor (without '#') as a string, or None if this item doesn't have an anchor.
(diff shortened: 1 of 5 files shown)


src/mkdocstrings/handlers/base.py
```diff
@@ -125,17 +125,14 @@ def render(self, data: CollectorItem, config: dict) -> str:
             The rendered template as HTML.
         """  # noqa: DAR202 (excess return section)
 
-    def get_anchor(self, data: CollectorItem) -> Optional[str]:
```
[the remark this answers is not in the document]
pawamoy: Sure, I'll update.

src/mkdocstrings/handlers/python.py
```diff
@@ -95,8 +95,11 @@ def render(self, data: CollectorItem, config: dict) -> str:  # noqa: D102 (ignor
             **{"config": final_config, data["category"]: data, "heading_level": heading_level, "root": True},
         )
 
-    def get_anchor(self, data: CollectorItem) -> str:  # noqa: D102 (ignore missing docstring)
-        return data.get("path")
+    def get_anchors(self, data: CollectorItem) -> Sequence[str]:  # noqa: D102 (ignore missing docstring)
+        try:
+            return (data["path"],)
+        except KeyError:
+            return ()
```
oprypin, line 102:
```suggestion
        return data.get("path", ())
```

because exceptions are costly

oprypin (dismissed):

src/mkdocstrings/handlers/python.py
pawamoy: But that would return the path directly, without being wrapped in a tuple :confused:

oprypin: Oops nevermind then

PR #355, opened 2021-12-28
feat: Support handlers spanning multiple locations

pawamoy:
Namespace packages can be installed into multiple locations (and not merged into a single one).
To find the right templates directory, we must therefore find the right handler path.

### DIFF
--- a/src/mkdocstrings/handlers/base.py
+++ b/src/mkdocstrings/handlers/base.py
@@ -9,6 +9,7 @@ It also provides two methods:
 """
 
 import importlib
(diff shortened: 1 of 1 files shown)


oprypin (dismissed):

pawamoy:
Arg, not sure what's up with Python 3.11, it's passing fine locally. I'll reinstall it, maybe it broke in the meantime.

pawamoy:
Yeah OK 3.11.a03 crashes. I'll allow failures.

PR #356, opened 2021-12-28
refactor: Extract the Python handler into its own repository

pawamoy:
That was a bit of work :stuck_out_tongue: 
~~It needs #355 to support the install cache feature of PDM (resulting in multi-location namespace packages: handlers).~~ Merged


pawamoy:
Any idea why CI keeps expecting Python 3.6 results?

oprypin:
Probably something in https://github.com/mkdocstrings/mkdocstrings/settings/branches but I'm just guessing because I don't have access to that

pawamoy:
Yeah I think I saw settings related to CI some time ago but I can't manage to find them...

oprypin:
Well have you tried the exact link I just sent??

pawamoy:
I did but I missed the edit button :rofl: 
It's there indeed, thanks

pawamoy:
You're now an owner of the org by the way

oprypin:
Thanks much :grin:

oprypin:
@pawamoy I don't intend to review, you can merge. Maybe I'll look at everything just before release

oprypin:
Actually hmm tomorrow is the day that I have time

pawamoy:
Take your time, I'll probably merge tomorrow instead :slightly_smiling_face:

pyproject.toml
```diff
@@ -35,10 +35,14 @@ dependencies = [
...
     "mkdocs>=1.2",
     "mkdocs-autorefs>=0.3.1",
+    "mkdocstrings-python-legacy>=0.2",
     "pymdown-extensions>=6.3",
-    "pytkdocs>=0.14.0",
 ]
 
+[project.optional-dependencies]
+python-legacy = ["mkdocstrings-python-legacy>=0.2"]
+python = ["mkdocstrings-python>=0.5.1"]
```
oprypin, line 44: Why do you think these "extras" are necessary or preferred? Why not just tell people to depend on `mkdocstrings-python` rather than `mkdocstrings[python]`? This makes the playing field more fair for new handlers which wouldn't have to be registered here.
Or I don't know...
Maybe it *is* nice, but anyway pointing this out, just in case you didn't realize that this isn't necessary.

oprypin:
@pawamoy OK you can consider everything approved, other than the comments I left :)

pyproject.toml
pawamoy: I know it's not necessary :slightly_smiling_face: 
And I do feel it's nice to have. Example: if you have both Crystal and Python code, you can depend on `mkdocstrings[crystal,python]`.
And I realize we should definitely add the `crystal` extra :smile: 

To me these extras are just a way to advertise some handlers as the *official* ones (could be community contributed handlers, just like yours, handlers from other GitHub accounts, etc.). Any handler can still be used by depending on it directly.

pawamoy: OK I'll go ahead and keep the extras, I like the idea, and I don't see any downside (apart from maintenance maybe which I'm willing to take on), let see how it unfolds :slightly_smiling_face:

PR #367, opened 2022-02-01
Support new mkdocstrings_handlers namespace

pawamoy:


pawamoy:
Let's not emit deprecation warnings for our own legacy handler. Users don't care about them.

pawamoy:
@oprypin do you want us not to emit these deprecation warnings for the crystal handler as well?

oprypin:
@pawamoy Well yes but that will just be done by making the handler use the new approach

pawamoy:
Indeed, it's just in case you don't plan to upgrade your handler immediately and don't want to bother your users with a deprecation warning they have no control on. Let me know.

docs/handlers/overview.md
```diff
@@ -42,7 +40,7 @@ See the documentation for
 [`BaseRenderer`][mkdocstrings.handlers.base.BaseRenderer].
 
 Check out how the
-[Python handler](https://github.com/pawamoy/mkdocstrings/blob/master/src/mkdocstrings/handlers/python.py)
+[Python handler](https://github.com/mkdocstrings/python/blob/master/src/mkdocstrings_handlers/python)
```
oprypin, line 43: https://github.com/mkdocstrings/mkdocstrings/pull/356 will remove this destination though

pawamoy: Hmmm no I think it's right: this link points to the extracted Python handler, i.e. its own repository in the org, and not mkdocstrings' repo.

PR #368, opened 2022-02-01
feat: Add `strip_p` filter to remove `p` tag around text

pawamoy:
Sometimes we convert a single line of Markdown and we don't want to keep the surrounding `<p>` tags. This filter allows to strip them.

Example use-case:

```html+jinja
<details class="{{ section.value.kind }}">
  <summary>{{ section.title|convert_markdown(heading_level, html_id)|strip_p }}</summary>
  {{ section.value.contents|convert_markdown(heading_level, html_id) }}
</details>
```

The filter does nothing if the markup is not wrapped in `<p>`.

### DIFF
--- a/src/mkdocstrings/handlers/base.py
+++ b/src/mkdocstrings/handlers/base.py
@@ -56,6 +56,24 @@ def do_any(seq: Sequence, attribute: str = None) -> bool:
     return any(_[attribute] for _ in seq)
 
 
+def do_strip_p(html: str) -> str:
+    """Strip `<p>` and `</p>` from the beginning and end of the given HTML text.
+
+    This is only useful for Markdown text that was rendered as a single paragraph.
+
+    Parameters:
+        html: The HTML text to strip from.
+
+    Returns:
+        The stripped HTML text.
+    """
+    if html.startswith("<p>"):
+        html = html[3:]
+    if html.endswith("</p>"):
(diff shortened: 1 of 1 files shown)


oprypin (changes requested):
Wait, that's not so nice to do :sweat_smile: 
I'll try if there's a nicer alternative

oprypin:
I have provided my attempt in https://github.com/mkdocstrings/mkdocstrings/pull/369. It's comparatively heavy, though :/

Which things I thought might be downsides and that I managed to resolve:

* I feel like the `convert_markdown` filter itself should be removing the tag, with a parameter. The current approach implies that there isn't any way to not have that tag there in the first place. But why (other than implementation details) should it be a separate step?

* What if the content is several paragraphs but it still tries to remove them? <code><del>&lt;p&gt;</del>Foo&lt;/p&gt;&lt;p&gt;Bar<del>&lt;/p&gt;</del></code>

* (Far-fetched but) what if the `p` tag has attributes? <code>&lt;p x="y"&gt;Bar<del>&lt;/p&gt;</del></code>
   Or generally this approach could somehow remove the start tag but not the end tag.

pawamoy:
> But why (other than implementation details) should it be a separate step?

No reason at all, I tried adding the option to the convert_markdown filter and it was ugly :sweat_smile:

PR #373, opened 2022-02-04
Fixed bad error catch

JoeHCQ1:
See text below for detailed explanation.

```python
        error = result.get("error")  # Recall that .get returns None if there is no error
        if error is not None:  # In my particular error case the error was an empty string, which failed the old if since an empty string ~= false.
            if "traceback" in result:
                error += f"\n{result['traceback']}"
            raise CollectionError(error)

        if "loading_errors" in result:  # Without this check, the for loop results in a KeyError
            for loading_error in result["loading_errors"]:
                log.warning(loading_error)

        if "parsing_errors" in result:  # Without this check, the for loop results in a KeyError
            for errors in result["parsing_errors"].values():
                for parsing_error in errors:
                    log.warning(parsing_error)

        # We always collect only one object at a time
        if "objects" not in result:  # This may be unnecessary, but since the error wasn't caught above, this also became a KeyError.
            raise CollectionError(f"No objects returned in {result} for python object: {identifier}")
        result = result["objects"][0]

```

### DIFF
--- a/src/mkdocstrings/handlers/python.py
+++ b/src/mkdocstrings/handlers/python.py
@@ -252,19 +252,23 @@ class PythonCollector(BaseCollector):
             raise CollectionError(error) from exception
 
         error = result.get("error")
-        if error:
+        if error is not None:
             if "traceback" in result:
                 error += f"\n{result['traceback']}"
             raise CollectionError(error)
 
-        for loading_error in result["loading_errors"]:
-            log.warning(loading_error)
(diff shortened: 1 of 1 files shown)


pawamoy:
Hi @JoeHCQ1, thanks a lot for your contribution :slightly_smiling_face: 
However could you open the PR against https://github.com/mkdocstrings/python-legacy instead? The Python handler was moved there.

PR #378, opened 2022-02-07
docs: Add recipes page

### LINKED ISSUE #128: In examples, ignore >>> and ... when selecting code
I would like to be able to take examples and paste them in a python interpreter

```python
class A:
    """"
    Test

    Example:
    >>> import time
    >>>
    >>> print('something')
    """"
```

In the rendered version of the class, I would like to omit the >>> and ... beginnings on the line as they break the Python interpreter.

This could be possible by using https://css-tricks.com/almanac/properties/u/user-select/

**Describe alternatives you've considered**
Are there other options?

**Additional context**

pawamoy:
I know the "Recipes" page can quickly become a mess, but I believe we can reorganize it continually as it grows.

Closes #128

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -1,10 +1,202 @@
 On this page you will find various recipes, tips and tricks
 for *mkdocstrings* and more generally Markdown documentation.
 
+## Automatic code reference pages
+
+*mkdocstrings* allows to inject documentation for any object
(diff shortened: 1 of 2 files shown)


pawamoy:
@oprypin, I wrote about your plugins in a second recipe, if you'd like to review :slightly_smiling_face:

pawamoy:
Merging now, we can always revise it later.

PR #390, opened 2022-02-23
Update python autodoc recipe

johnthagen:
- Fix types to match signature of `Nav`: https://github.com/oprypin/mkdocs-gen-files/blob/4d5d4ac497ef50c070253c1ae4062b395b03413c/mkdocs_gen_files/nav.py#L14
- Simplify file writing by using file handle directly

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -172,18 +172,18 @@ for path in sorted(Path("src").rglob("*.py")):
     doc_path = path.relative_to("src").with_suffix(".md")
     full_doc_path = Path("reference", doc_path)
 
-    parts = list(module_path.parts)
+    parts = tuple(module_path.parts)
 
     if parts[-1] == "__init__":
         parts = parts[:-1]
     elif parts[-1] == "__main__":
         continue
 
-    nav[parts] = doc_path  # (1)
+    nav[parts] = str(doc_path)  # (1)
 
     with mkdocs_gen_files.open(full_doc_path, "w") as fd:
         ident = ".".join(parts)
-        print("::: " + ident, file=fd)
+        fd.write(f":::{ident}")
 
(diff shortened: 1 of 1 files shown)


pawamoy:
Mypy was complaining, right? In that case, LGTM :+1:

docs/recipes.md
```diff
@@ -172,18 +172,18 @@ for path in sorted(Path("src").rglob("*.py")):
...
     elif parts[-1] == "__main__":
         continue
 
-    nav[parts] = doc_path  # (1)
+    nav[parts] = str(doc_path)  # (1)
 
     with mkdocs_gen_files.open(full_doc_path, "w") as fd:
         ident = ".".join(parts)
-        print("::: " + ident, file=fd)
+        fd.write(f":::{ident}")
```
pawamoy, line 186:
```suggestion
        fd.write(f"::: {ident}")
```

docs/recipes.md
```diff
@@ -252,11 +252,11 @@ for path in sorted(Path("src").rglob("*.py")):
     elif parts[-1] == "__main__":
         continue
 
-    nav[parts] = doc_path
+    nav[parts] = str(doc_path)
 
     with mkdocs_gen_files.open(full_doc_path, "w") as fd:
         ident = ".".join(parts)
-        print("::: " + ident, file=fd)
+        fd.write(f":::{ident}")
```
pawamoy, line 259:
```suggestion
        fd.write(f"::: {ident}")
```

johnthagen:
PyCharm actually caught this one, rather than `mypy` because upstream does not have `py.typed` needed for `mypy` to analyze it.

I opened an issue for that to get upstream, at which point, yes, `mypy` would have caught this.

- https://github.com/oprypin/mkdocs-gen-files/issues/6

PR #394, opened 2022-02-26
docs: Update note about supported docstring options

thatlittleboy:
Just a small update to the documentation, pointing out that docstring options are available now under the experimental python handler. :)


pawamoy (approved):
Nice, thanks!

PR #397, opened 2022-03-03
Prevent Nonetype error in call to Markdown.convert

JoeHCQ1:
When docstring style is set to numpy, if you are missing part of the expected docstring you'll get a Nonetype error when it is passed to the Markdown module method. I opened a [PR in Markdown](https://github.com/Python-Markdown/markdown/pull/1223) to add a None check to prevent the error but [they explained that the check should be in the caller](https://github.com/Python-Markdown/markdown/pull/1223#issuecomment-1050912432) so I'm opening this PR.

Here is the error caused by the code as is in 0.18.1:

```
  File "/home/me/repo/venv/lib/python3.8/site-packages/mkdocstrings/handlers/base.py", line 219, in do_convert_markdown
    return Markup(self._md.convert(text))
  File "/home/me/repo/venv/lib/python3.8/site-packages/markdown/core.py", line 248, in convert
    if not source.strip():
AttributeError: 'NoneType' object has no attribute 'strip'
```

I edited my local copy of the file to test my changes and it worked. Rather than crashing, it kept going. If it's of interest, here are the warnings it gave. I expect one of these caused the crash:

```
WARNING  -  mkdocstrings: somemodule.someclass.somefunca: No type annotation for parameter 'model'
WARNING  -  mkdocstrings: somemodule.someclass.somefuncb: Empty return description
WARNING  -  mkdocstrings: somemodule.someclass.somefuncc: Empty return description
WARNING  -  mkdocstrings: somemodule.someclass.somefuncd: Empty return description
```

For final added context, I got this crash when running `python -m mkdocs build`.

### DIFF
--- a/src/mkdocstrings/handlers/base.py
+++ b/src/mkdocstrings/handlers/base.py
@@ -215,16 +215,19 @@ class BaseRenderer(ABC):
         treeprocessors[HeadingShiftingTreeprocessor.name].shift_by = heading_level
         treeprocessors[IdPrependingTreeprocessor.name].id_prefix = html_id and html_id + "--"
         treeprocessors[ParagraphStrippingTreeprocessor.name].strip = strip_paragraph
-        try:
-            if text:
-                return Markup(self._md.convert(text))
-            else:
-                return Markup()
-        finally:
-            treeprocessors[HeadingShiftingTreeprocessor.name].shift_by = 0
-            treeprocessors[IdPrependingTreeprocessor.name].id_prefix = ""
-            treeprocessors[ParagraphStrippingTreeprocessor.name].strip = False
-            self._md.reset()
+        
+        html_str = Markup()
+        if text:
+            html_str = Markup(self._md.convert(text))
+        else:
+            log.warning(f"Mising expected text in {html_id}")
+           
+        treeprocessors[HeadingShiftingTreeprocessor.name].shift_by = 0
+        treeprocessors[IdPrependingTreeprocessor.name].id_prefix = ""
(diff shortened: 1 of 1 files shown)


JoeHCQ1:
I've since played with it more. The errors I listed in the description definitely didn't cause this. I'm not sure what kind of error is causing the None being passed into Markdown.

Doing some more testing, it looks like it happens when you don't add a description to a parameters - example:

```python
"""
My numpydoc string documented function foo

Parameters
------------
param : str  # if I don't add a description below I get that None error
    Description to prevent Nonetype error.
"""

Returns
--------
ret : str  # Interestingly if I omit elements of this I get specific warnings telling me what is wrong
    # I can also omit the description and get a pointed warning but no None error.
```

JoeHCQ1:
I'm not real sure how all the pieces fit together, but it looks like this is where we're missing numpy docstring validation: https://github.com/mkdocstrings/pytkdocs/blob/master/src/pytkdocs/parsers/docstrings/numpy.py#L104

it should have an additional check like the return does here: https://github.com/mkdocstrings/pytkdocs/blob/master/src/pytkdocs/parsers/docstrings/numpy.py#L196 - I think that means we need a new PR - although this catch here may be a good idea still since it keeps the warning but prevents the crash.

JoeHCQ1:
Closing this in preference for https://github.com/mkdocstrings/pytkdocs/pull/138

pawamoy:
Thank you very much for your thorough investigation and for coming back and closing the PR in favor of the pytkdocs one :slightly_smiling_face:

Issue #364, opened 2022-01-17
New Python handler is here


pawamoy:
Hello everyone!

I've been working on a new Python handler. This issue serves as a public TODO list for what is needed before getting it out of the "experimental" state.

The new handler is [mkdocstrings/python](https://github.com/mkdocstrings/python), and the old one has been extracted into [mkdocstrings/python-legacy](https://github.com/mkdocstrings/python-legacy). The new data collector, [mkdocstrings/griffe](https://github.com/mkdocstrings/griffe), is an improved version of [mkdocstrings/pytkdocs](https://github.com/mkdocstrings/pytkdocs). It's able to both visit (AST) and inspect (exec + introspection) objects, and has an overall better design.

### Phase 1

- [x] Finish implementing support for namespace packages in griffe
- [x] Publish a new griffe release
- [x] Fix docstrings admonitions rendering in mkdocstrings/python
- [x] Use the new `mkdocstrings_handlers` namespace in mkdocstrings/python
- [x] Publish a new mkdocstrings/python release
- [x] Open mkdocstrings PR to allow it to load handlers from the `mkdocstrings_handlers` namespace (with backward-compatibility and deprecation warnings) #356 and #367
- [x] Update PR #356 with the `python` experimental extra actually installing mkdocstrings-python
- [x] Update docs about the new handler: things that can possibly stop working (dynamic objects, custom templates?), deprecations, user warnings, how to ignore them, etc. #371 
- [x] Add user warning (specify extra) when using mkdocstrings/python-legacy
- [x] Publish new legacy handler release
- [x] Merge all mkdocstrings PRs
- [x] Publish a new mkdocstrings release 
- [x] At this point users can start testing the new handler by depending on `mkdocstrings[python]`
- [x] Add back usage docs on the legacy handler docs
- [x] Enjoy a round of bugtracker cleaning :drooling_face: 

---

### Phase 2

- [x] Stop depending on the legacy handler in mkdocstrings
- [x] Remove user warning when using mkdocstrings/python-legacy (since by now the extra is mandatory)
- [x] Use the new mkdocstrings_handlers namespace in mkdocstrings/python-legacy as well
- [x] Refactor mkdocstrings to get rid of the `BaseCollector` and `BaseRenderer` classes: merge everything into the `BaseHandler` class #413 
- [x] Stop splitting options into `selection` and `rendering` keys. Keep backward compatibility (with deprecation warnings). #420
[...]

PR #420, opened 2022-04-10
refactor: Deprecate 'selection' and 'rendering' YAML keys

pawamoy:
Commit message:

As seen in commit eb822cb, the separation of
the selection/collection and rendering configuration
is problematic. This commit further reduces this
separation by merging the two current YAML keys
'selection' and 'rendering' into a single
'options' key. It means both the 'collect' and
'render' methods have access to every option.

The old keys can still be used, but are deprecated,
and therefore a deprecation warning is emitted
if these keys are found in either global or local
configurations.

---

This was mentioned in #364.
Example diffs:

```diff
 # mkdocs.yml
 plugins:
 - mkdocstrings:
     handlers:
       python:
-        selection:
+        options:
           docstring_style: google
           docstring_options:
             ignore_init_summary: yes
-        rendering:
           merge_init_into_class: yes
           show_submodules: no
```

```diff
 # My Class

 ::: my_package.MyClass
-    rendering:
+    options:
       show_root_toc_entry: no
```

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -91,11 +91,10 @@ plugins:
         - https://docs.python.org/3/objects.inv
         - https://docs.python-requests.org/en/master/objects.inv  # demonstration purpose in the docs
         - https://mkdocstrings.github.io/autorefs/objects.inv
-        selection:
+        options:
           docstring_style: google
           docstring_options:
             ignore_init_summary: yes
-        rendering:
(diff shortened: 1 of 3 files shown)


pawamoy:
I intend to merge this in two days :slightly_smiling_face:

pawamoy:
Merging, we can always revise the changes before releasing them.

PR #429, opened 2022-05-08
docs: Update docs

pawamoy:
Hey @oprypin (and other watchers), I've updated the docs for the upcoming version (0.19).
I'll add comments on interesting points.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,21 +14,20 @@ cd mkdocstrings
 make setup
 ```
 
-!!! note
-    If it fails for some reason,
-    you'll need to install
-    [PDM](https://github.com/pdm-project/pdm)
-    manually.
-
-    You can install it with:
-
-    ```bash
-    python3 -m pip install --user pipx
-    pipx install pdm
-    ```
-
-    Now you can try running `make setup` again,
-    or simply `pdm install`.
+> NOTE: If it fails for some reason,
+> you'll need to install
+> [PDM](https://github.com/pdm-project/pdm)
+> manually.
+> 
+> You can install it with:
+> 
+> ```bash
+> python3 -m pip install --user pipx
+> pipx install pdm
+> ```
+> 
+> Now you can try running `make setup` again,
+> or simply `pdm install`.
 
 You now have the dependencies installed.
 
--- a/docs/handlers/overview.md
+++ b/docs/handlers/overview.md
@@ -172,7 +172,7 @@ using the Jinja templates provided by your package.
 You must implement a `get_handler` method at the module level.
 This function takes the following parameters:
 
--  `theme` (string, theme name),
+- `theme` (string, theme name)
 - `custom_templates` (optional string, path to custom templates directory)
 - `config_file_path` (optional string, path to the config file)
 
--- a/docs/handlers/overview.md
+++ b/docs/handlers/overview.md
@@ -14,9 +14,8 @@ Since version 0.18, a new, experimental Python handler is available.
 It is based on [Griffe](https://github.com/mkdocstrings/griffe),
 which is an improved version of [pytkdocs](https://github.com/mkdocstrings/pytkdocs).
 
-Note that the experimental handler does not yet offer the same features as the legacy one.
-If you are making extensive use of the current (legacy) Python handler selection and rendering options,
-you might want to wait a bit before trying the experimental handler.
+Note that the experimental handler does not yet support third-party libraries
+like Django, Marshmallow, Pydantic, etc.
 It is also not completely ready to handle dynamically built objects,
 like classes built with a call to `type(...)`.
 For most other cases, the experimental handler will work just fine.
@@ -25,7 +24,7 @@ If you want to keep using the legacy handler as long as possible,
 you can depend on `mkdocstrings-python-legacy` directly,
 or specify the `python-legacy` extra when depending on *mkdocstrings*:
 
-```toml
+```toml title="pyproject.toml"
 # PEP 621 dependencies declaration
 # adapt to your dependencies manager
 [project]
@@ -37,25 +36,13 @@ dependencies = [
 The legacy handler will continue to "work" for many releases,
 as long as the new handler does not cover all previous use-cases.
 
-Using the legacy handler will emit a `UserWarning` stating that users
-should specify the `python-legacy` extra when depending on *mkdocstrings*.
-The warning will be emitted even if you do specify the extra, as we have
-no way to detect it.
-
-Warnings can be globally ignored by setting the
-[`PYTHONWARNINGS` environment variable](https://docs.python.org/3/library/warnings.html#describing-warning-filters):
-
-```bash
-PYTHONWARNINGS=ignore::UserWarning:mkdocstrings.handlers.python
-```
-
 ### Migrate to the experimental Python handler
 
 To use the new, experimental Python handler,
 you can depend on `mkdocstrings-python` directly,
 or specify the `python` extra when depending on *mkdocstrings*:
 
-```toml
+```toml title="pyproject.toml"
 # PEP 621 dependencies declaration
 # adapt to your dependencies manager
 [project]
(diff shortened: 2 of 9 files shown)


CONTRIBUTING.md
```diff
@@ -14,21 +14,20 @@ cd mkdocstrings
...
+> 
+> You can install it with:
+> 
+> ```bash
+> python3 -m pip install --user pipx
+> pipx install pdm
+> ```
+> 
+> Now you can try running `make setup` again,
+> or simply `pdm install`.
```
pawamoy, line 30: Switched to callouts :heart: 
This one was important to change since CONTRIBUTING is often read on GitHub directly.
I did not switch all admonitions to callouts, because the markup tends to be heavy, but I'm not opposed to a full switch later.

docs/handlers/overview.md
```diff
@@ -37,25 +36,13 @@ dependencies = [
...
-The warning will be emitted even if you do specify the extra, as we have
-no way to detect it.
-
-Warnings can be globally ignored by setting the
-[`PYTHONWARNINGS` environment variable](https://docs.python.org/3/library/warnings.html#describing-warning-filters):
-
-```bash
-PYTHONWARNINGS=ignore::UserWarning:mkdocstrings.handlers.python
-```
-
```
pawamoy, line 51: Next version doesn't emit the warning anymore.

docs/handlers/overview.md
```diff
@@ -66,107 +53,94 @@ dependencies = [
...
-```jinja title="templates/theme/docstring.html"
-{% extends "_base/docstring.html" %}
-{% block contents %}
-  {{ block.super }}
-  Additional contents
-{% endblock contents %}
-```
-
-**Block-level customization is not ready yet.**
```
pawamoy, line 162: These paragraphs were moved in the handler's docs.

docs/handlers/overview.md
```diff
@@ -182,44 +156,51 @@ For *mkdocstrings*, a custom handler package would have the following structure:
...
+using the Jinja templates provided by your package.
 
-See the documentation for
-[`BaseHandler`][mkdocstrings.handlers.base.BaseHandler],
-[`BaseCollector`][mkdocstrings.handlers.base.BaseCollector] and
-[`BaseRenderer`][mkdocstrings.handlers.base.BaseRenderer].
+You must implement a `get_handler` method at the module level.
+This function takes the following parameters:
+
+-  `theme` (string, theme name),
```
pawamoy, line 175:
```suggestion
- `theme` (string, theme name)
```

docs/recipes.md
```diff
@@ -283,19 +283,6 @@ plugins:
...
-
-    ```yaml title="mkdocs.yml" hl_lines="8"
-    plugins:
-    - mkdocstrings:
-        handlers:
-          python:
-            rendering:
-              show_submodules: no
-    ```
-
```
pawamoy, line 298: `show_submodules` changed to `False` by default in the next handler version.

docs/theming.md
```diff
@@ -21,59 +21,47 @@ To use custom templates and override the theme ones,
...
-    - `method.html`: to render methods
-    - `module.html`: to render modules
-- `docstring.html`: to render docstrings
-    - `attributes.html`: to render attributes sections of docstrings
-    - `examples.html`: to render examples sections of docstrings
-    - `exceptions.html`: to render exceptions/"raises" sections of docstrings
-    - `parameters.html`: to render parameters/arguments sections of docstrings
-    - `return.html`: to render "return" sections of docstrings
-- `properties.html`: to render the properties of an object (`staticmethod`, `read-only`, etc.)
-- `signature.html`: to render functions and methods signatures
```
pawamoy, line 76: This is specific to the Python handler and Material theme, nothing to do here.

docs/theming.md
```diff
@@ -21,59 +21,47 @@ To use custom templates and override the theme ones,
...
+to modify small part of the templates without copy-pasting the whole files.
+
+See the documentation about templates for:
+
+- the Python handler: https://mkdocstrings.github.io/python/customization/#templates
```
pawamoy, line 64: Let me know if we can link to the Crystal handler's docs as well.

docs/theming.md
```diff
@@ -89,53 +77,18 @@ Every template has access to a `log` function, allowing to log messages as usual
...
-            <small class="prop" style="background-color: red; color: white !important;">private</small>
-            <small class="prop" style="background-color: green; color: white !important;">property</small>
-            <small class="prop" style="background-color: yellow; color: black !important;">read-only</small>
-        </span></h3>
-
-        As you can see, CSS is not my field of predilection...
+Since each handler provides its own set of templates, with their own CSS classes,
+we cannot list them all here. See the documentation about CSS classes for:
+
+- the Python handler: https://mkdocstrings.github.io/python/customization/#css-classes
```
pawamoy, line 83: As stated, these were specific to the Python handler (moved there). Again, let me know if we can link to the Crystal handler's docs.

docs/troubleshooting.md
```diff
@@ -115,23 +114,17 @@ use this workaround.
...
 ## The generated documentation does not look good
 
-Are you using the Material theme?
-
-- "No": We do not support any other theme yet.
-  Check the [bugtracker][bugtracker] to see if there is a feature request
-  asking to support your theme. If you find one, vote with a thumbs up. If not, you can open a ticket.
-- "Yes": Please open an ticket on the [bugtracker][bugtracker] with a detailed
-  explanation and screenshots of the bad-looking parts.
-
```
pawamoy, line 125: This was outdated.

mkdocs.yml
```diff
@@ -32,8 +33,7 @@ nav:
 
 theme:
   name: material
-  icon:
-    logo: material/currency-sign
+  logo: logo.png
```
pawamoy, line 36: Thank you so much for the logo @oprypin :rocket:

pawamoy:
Now using the SVG logo, thanks

docs/theming.md
oprypin: Sure, you can link to https://mkdocstrings.github.io/crystal/styling.html

docs/theming.md
oprypin: Feel free to link to https://mkdocstrings.github.io/crystal/styling.html#custom-styles

docs/usage.md
```diff
@@ -47,11 +41,10 @@ Check the documentation for your handler of interest in [Handlers](handlers/over
 
         ::: my_package.my_module.MyClass
             handler: python
-            selection:
+            options:
```
oprypin, line 44: Mmm it's probably too late (and I'm commenting on implementation, not on the docs), but it sure would be nice to just not require the redundant `options` key; it's almost always going to be just this one key with its sub-keys. `handler` could just be reserved and popped before passing it.

pawamoy: We still need to be able to differentiate the global-only options of the handler and the ones that can be overridden locally, but that would work for local options yes. Maybe this would feel a bit inconsistent though.

pyproject.toml
```diff
@@ -65,6 +65,7 @@ editable-backend = "editables"
 [tool.pdm.dev-dependencies]
 duty = ["duty>=0.7"]
 docs = [
+    "markdown-callouts>=0.2.0",
```
oprypin, line 68:
```suggestion
    "markdown-callouts>=0.2.0",
    "mkdocs>=1.3",
```

oprypin (approved):

pyproject.toml
pawamoy: It wouldn't hurt, but MkDocs is already a prod-dependency so is this really useful? I personally never install the docs group without the default one.

pawamoy: Ah, yeah, makes sense. Thanks.

docs/usage.md
oprypin: Yeah I don't know... at first I was thinking that maybe it would be good to eliminate that distinction even for the main config. But that'd be too hacky, basically validation for the constructor's arguments would have to be dropped.

https://github.com/mkdocstrings/mkdocstrings/blob/5867162b2eebf4829d39b6fe74843b60ced05a30/docs/handlers/overview.md?plain=1#L200-L205

For reference, another config keys documentation https://mkdocstrings.github.io/crystal/configuration.html

pawamoy: What if we allow both ways for local options? 

```
::: id
    options:
      foo: bar
```

```
::: id
    foo: bar
```

And if you want to change the handler, you must use the options key:

```
::: id
    handler: qux
    options:
      foo: bar
```

pawamoy: Well, lets keep the options key mandatory for now. We can always change things later.

PR #443, opened 2022-06-14
Fix edit link for autogenerated docs

davfsa:
Been going crazy about this for the last couple of hours and unable to figure out for my personal docs and saw that this behaviour was also reproducible in the repo. Eventually found it and it is caused by the use of [urllib.parse.urljoin](https://github.com/oprypin/mkdocs-gen-files/blob/master/mkdocs_gen_files/plugin.py#L49-L51) in the gen_files plugin, which misinterprets the links if they are not ended in `/` (@oprypin would you rather this gets fixed there instead?)

Additionally, I noticed that the `edit_uri` makes the edit links for the now correct links have `docs/` as a base, which makes it link incorrectly, so I added a bit of a hacky fix to avoid that

An example of the now broken page links can be seen [here](https://mkdocstrings.github.io/reference/extension/), where clicking on edit page leads to a 404.

```diff
- https://github.com/mkdocstrings/blob/master/docs/src/mkdocstrings/extension.py
+ https://github.com/mkdocstrings/mkdocstrings/blob/master/docs/src/mkdocstrings/extension.py
```

EDIT: Would also like to add that both these issues can be seen across all other repositories for extensions for mkdocstrings

### DIFF
--- a/docs/gen_ref_nav.py
+++ b/docs/gen_ref_nav.py
@@ -24,7 +24,7 @@ for path in sorted(Path("src").glob("**/*.py")):
         ident = ".".join(parts)
         print("::: " + ident, file=fd)
 
-    mkdocs_gen_files.set_edit_path(full_doc_path, path)
+    mkdocs_gen_files.set_edit_path(full_doc_path, Path("../") / path)
 
 with mkdocs_gen_files.open("reference/SUMMARY.md", "w") as nav_file:
     nav_file.writelines(nav.build_literate_nav())
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -21,11 +21,11 @@ which live in the `src` folder:
 📁 repo
 └─╴📁 src
     └─╴📁 project
-        ├─╴📄 lorem.py
-        ├─╴📄 ipsum.py
-        ├─╴📄 dolor.py
-        ├─╴📄 sit.py
-        └─╴📄 amet.py
+        ├─╴📄 lorem
+        ├─╴📄 ipsum
+        ├─╴📄 dolor
+        ├─╴📄 sit
+        └─╴📄 amet
 ```
 
 Without an automatic process, you will have to manually
@@ -108,11 +108,7 @@ for path in sorted(Path("src").rglob("*.py")):  # (1)
 9. We can even set the `edit_uri` on the pages.
 
 > NOTE:
-> It is important to look out for correct edit page behaviour when using generated pages,
-> as `full_path` will be relative to the location of your `mkdocs.yml` path. You can fix
-> incorrect linking by traversing tree and adding it to the path before passing
-> it to `set_edit_path`.
->
+> It is important to look out for correct edit page behaviour when using generated pages.
 > For example, if we have `edit_uri` set to `blob/master/docs/` and the following
 > file structure:
 >
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -21,11 +21,11 @@ which live in the `src` folder:
 📁 repo
 └─╴📁 src
     └─╴📁 project
-        ├─╴📄 lorem
-        ├─╴📄 ipsum
-        ├─╴📄 dolor
-        ├─╴📄 sit
-        └─╴📄 amet
+        ├─╴📄 lorem.py
+        ├─╴📄 ipsum.py
+        ├─╴📄 dolor.py
+        ├─╴📄 sit.py
+        └─╴📄 amet.py
 ```
 
 Without an automatic process, you will have to manually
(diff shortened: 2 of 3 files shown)


mkdocs.yml
```diff
@@ -1,7 +1,7 @@
 site_name: "mkdocstrings"
 site_description: "Automatic documentation from sources, for MkDocs."
 site_url: "https://mkdocstrings.github.io/"
-repo_url: "https://github.com/mkdocstrings/mkdocstrings"
+repo_url: "https://github.com/mkdocstrings/mkdocstrings/"
```
oprypin, line 4: Thanks. Yes, this is definitely a bug in mkdocs_gen_files; I had too much faith in `urljoin` :D
mkdocs-gen-files should gain these exact lines:
https://github.com/mkdocs/mkdocs/blame/17ff566d2a91703396e954efc521d3869472ba79/mkdocs/structure/pages.py#L108-L109
Ah and it's possible to write tests for that too, but I don't have anything close to it there :(

docs/gen_ref_nav.py
```diff
@@ -24,7 +24,7 @@
         ident = ".".join(parts)
         print("::: " + ident, file=fd)
 
-    mkdocs_gen_files.set_edit_path(full_doc_path, path)
+    mkdocs_gen_files.set_edit_path(full_doc_path, Path("../") / path)
```
oprypin, line 27: I did not foresee the possibility that people would need to use `..` *and* even more so I'm surprised that it actually works very cleanly.
```pycon
>>> urllib.parse.urljoin('http://example.com/foo/bar/tree/master/docs/', '../src/stuff/hi.py')
'http://example.com/foo/bar/tree/master/src/stuff/hi.py'
```

oprypin (commented):
I think

docs/gen_ref_nav.py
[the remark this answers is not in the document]
davfsa: I did not update that one because we never set edit_uri anywhere, so it might be confusing if they do and then it starts linking wrongly. I will add a note about it tho

docs/recipes.md
```diff
@@ -107,6 +107,37 @@ for path in sorted(Path("src").rglob("*.py")):  # (1)
 8. Actually write to the magic file.
 9. We can even set the `edit_uri` on the pages.
 
+> NOTE:
+> If your `edit_uri` is the default value or you have overwritten, you might
```
pawamoy, line 111:
```suggestion
> If your `edit_uri` is the default value or if you have overwritten it, you might
```

docs/recipes.md
```diff
@@ -107,6 +107,37 @@ for path in sorted(Path("src").rglob("*.py")):  # (1)
...
+> have to traverse the tree until you find the correct file from your `edit_uri`.
```
pawamoy, line 112:
```suggestion
> have to traverse the tree until you find the correct file for your `edit_uri`.
```

?

pawamoy (approved):
Thanks!

pawamoy:
Sorry for the troublesome hours :sweat_smile: And thank you so much for investigating and fixing this!

docs/recipes.md
pawamoy: By the way, these two cases represent... all the cases? So maybe the first sentence (condition) could be dropped?

davfsa: When writing it I had in mind that you have to look out for it in both cases, didnt read back and realize that it covers all conditions, will change

davfsa:
> Sorry for the troublesome hours 😅 And thank you so much for investigating and fixing this!

Dont worry, different timezone, so I just woke up :)

docs/recipes.md
davfsa: Rewrote the paragraph to be more clear :)

docs/recipes.md
```diff
@@ -21,11 +21,11 @@ which live in the `src` folder:
...
-        ├─╴📄 lorem
-        ├─╴📄 ipsum
-        ├─╴📄 dolor
-        ├─╴📄 sit
-        └─╴📄 amet
+        ├─╴📄 lorem.py
+        ├─╴📄 ipsum.py
+        ├─╴📄 dolor.py
+        ├─╴📄 sit.py
+        └─╴📄 amet.py
```
pawamoy, line 28: I had intentionally left out the extensions since mkdocstrings is language agnostic. I'd prefer you send another PR with that change if you think it improves the docs :slightly_smiling_face:

docs/recipes.md
```diff
@@ -107,6 +107,44 @@ for path in sorted(Path("src").rglob("*.py")):  # (1)
 8. Actually write to the magic file.
 9. We can even set the `edit_uri` on the pages.
 
+> NOTE:
+> It is important to look out for correct edit page behaviour when using generated pages,
+> as `full_path` will be relative to the location of your `mkdocs.yml` path. You can fix
```
pawamoy, line 112:
```suggestion
> as `full_doc_path` will be relative to your docs directory. You can fix
```

docs/recipes.md
```diff
@@ -107,6 +107,44 @@ for path in sorted(Path("src").rglob("*.py")):  # (1)
...
+> incorrect linking by traversing tree and adding it to the path before passing
+> it to `set_edit_path`.
+>
+> For example, if we have `edit_uri` set to `blob/master/docs/` and the following
+> file structure:
```
pawamoy, line 117: I find the first paragraph confusing. The second one is better as it shows rather than telling. So I suggest simplifying it like so:

```suggestion
> It is important to look out for correct edit page behaviour when using generated pages.
> For example, if we have `edit_uri` set to `blob/master/docs/` and the following
> file structure:
```

docs/recipes.md
davfsa: It didnt hit at the time that it is, will revert 😅
```suggestion
        ├─╴📄 lorem
        ├─╴📄 ipsum
        ├─╴📄 dolor
        ├─╴📄 sit
        └─╴📄 amet
```

docs/recipes.md
```diff
@@ -107,6 +107,40 @@ for path in sorted(Path("src").rglob("*.py")):  # (1)
...
+> It is important to look out for correct edit page behaviour when using generated pages.
+> For example, if we have `edit_uri` set to `blob/master/docs/` and the following
+> file structure:
+>
+> ```
+> 📁 repo
+> ├─ 📄 mkdocs.yml
+> │
+> ├─ 📁 docs
+> │   ├─╴index.md
```
pawamoy, line 120:
```suggestion
> │   ├─╴📄 index.md
```

pawamoy (approved):
OK, LGTM, thanks a lot!

Hnasar:
I had to rediscover this and i was about to open a similar merge request! Can you deploy the updated docs @pawamoy ? The live site still has the old version https://mkdocstrings.github.io/recipes/

pawamoy:
Thanks for the reminder, I'll deploy the new docs :slightly_smiling_face:

PR #469, opened 2022-08-18
feat: Add syntax to nest autodocs into other autodocs

oprypin:
The syntax:
```markdown
::: foo --------

This text will be treated as an appendix to whatever content the docstring of `foo` produced

Identifiers can be nested too:

::: foo.Bar

Still going

--------

Now we are outside of the fence
```

The beginning fence exactly shares syntax with the syntax for `<hr>` and then the ending fence has to exactly match it.

https://user-images.githubusercontent.com/371383/185492525-987cbe8e-5088-4dab-b46c-9a7d0402f81b.mp4

https://user-images.githubusercontent.com/371383/185492513-e3e4517e-316b-4600-b058-a2bc8a08e417.mp4

---

This will be really useful for things like https://github.com/mkdocs/mkdocs/pull/2934. I've had a need for this feature for a long time for my other projects too. Only now came up with a decent syntax idea for it.

Please review commit-by-commit.

I am giving up on the "quality" lint complaints for now, they are totally unreasonable.

### DIFF
--- a/src/mkdocstrings/extension.py
+++ b/src/mkdocstrings/extension.py
@@ -31,7 +31,7 @@ import yaml
 from jinja2.exceptions import TemplateNotFound
 from markdown import Markdown
 from markdown.blockparser import BlockParser
-from markdown.blockprocessors import BlockProcessor
+from markdown.blockprocessors import BlockProcessor, HRProcessor
 from markdown.extensions import Extension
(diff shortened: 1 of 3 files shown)


pawamoy:
@oprypin this can be achieved with the new pymdownx blocks right? Should we close this PR? I plan on (re)implementing the extension as a block anyway (not sure when though).

PR #474, opened 2022-09-04
docs: Clarify `custom_templates` folder location in options documentation

KSneijders:
A small PR to (in my opinion) improve the clarity of the `custom_templates` config option.  
The reason for this is that when I read the current line:

> the path to a directory containing custom templates. The path is relative to the docs directory

I assumed it meant it was relative to **within** the docs directory. 
Only after looking up other repositories with examples I noticed that you have to include `docs` in the path if your templates are in that folder.

This is not meant to be final, just an example. If this is not the place for the explanation, I don't mind moving it either.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -108,7 +108,8 @@ The above is equivalent to:
 
 - `default_handler`: the handler that is used by default when no handler is specified.
 - `custom_templates`: the path to a directory containing custom templates.
-  The path is relative to the docs directory.
+  The path is relative to the location of the docs directory.
+  If your templates folder is insde your docs directory, you do: `docs/...`.
   See [Theming](theming.md).
 - `handlers`: the handlers global configuration.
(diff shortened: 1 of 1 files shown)


pawamoy:
Hi @KSneijders, thank you for your help on improving the docs!

Actually, it seems the `custom_templates` value is relative to the *current working directory*, which should be fixed to make it relative to the `mkdocs.yml` file instead. I'll open a new issue.

pawamoy:
You can update this PR to say that the custom templates folder is relative to the current working directory instead of the docs one.

KSneijders:
Hey @pawamoy!
Thanks for the response! It being relative to the CWD explains a lot.  
I've changed the explanation to a simple line without example as the CWD could be different for everyone which makes it difficult to give examples.

pawamoy:
Thank you for the update, LGTM, merging :slightly_smiling_face:

PR #504, opened 2023-01-05
feat: add plugin `enabled` config option

### LINKED ISSUE #478: Add option to disable plugin in mkdocs.yml
When editing sections of my docs that are not using mkdocstrings, the live preview takes a while to update because it will run mkdocstrings on every change. 

A config parameter in the plugin configuration section of the `mkdocs.yml` to disable the plugin. Together with the [`!ENV`](https://www.mkdocs.org/user-guide/configuration/#environment-variables) functionality of mkdocs it makes it easy to disable the plugin without having to modify the `mkdocs.yml`. So it could look like:

```yaml
plugins:
- mkdocstrings:
    disabled: !ENV [DISABLE_MKDOCSTRING, False]
    default_handler: python
    handlers:
      python:
        options:
          show_source: false

```

I use another plugin [`mkdocs-simple-hooks`](https://github.com/aklajnert/mkdocs-simple-hooks#disabling-the-plugin) which uses this  pattern. Here the config option is `enabled` and defaults to `True`. I think it makes more sense to have it as `disabled` and default to `False` as you as default has it enabled and then can use this option if you want to disable it.

**Describe alternatives you've considered**
- Comment out the mkdocstrings section of the `mkdocs.yml`. Works, but is annoying and I will accidentally commit the change when I run `git add -u`.
- unwatch the sections of the docs which uses mkdocstrings. Not sure if it's possible to unwatch parts of the doc folder? In any case it would still be a problem when editing files that uses mkdocstrings but you are not focusing on that.  

**Additional context**
I can make the PR if you think this change makes sense.

StefanBRas:
Fixes #478.

For the docs entry I guessed that it will be released in version 0.20, this might have to be adjusted.

I considered adding a test but I couldn't see some obvious way to extend the test suite.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -113,6 +113,9 @@ The above is equivalent to:
 - `handlers`: the handlers global configuration.
 - `enable_inventory`: whether to enable inventory file generation.
   See [Cross-references to other projects / inventories](#cross-references-to-other-projects-inventories)
+- `enabled` **(New in version 0.20)**: Whether to enable the plugin. Defaults to `true`.
+  Can be used to reduce build times when doing local development.
+  Especially useful when used with environment variables (see example below).
 - `watch` **(deprecated)**: a list of directories to watch while serving the documentation.
   See [Watch directories](#watch-directories). **Deprecated in favor of the now built-in
   [`watch` feature of MkDocs](https://www.mkdocs.org/user-guide/configuration/#watch).
@@ -121,6 +124,7 @@ The above is equivalent to:
     ```yaml title="mkdocs.yml"
     plugins:
     - mkdocstrings:
+        enabled: !ENV [ENABLE_MKDOCSTRINGS, true]
         custom_templates: templates
         default_handler: python
         handlers:
@@ -357,3 +361,4 @@ For example, it will not tell the Python handler to look for packages in these p
 (the paths are not added to the `PYTHONPATH` variable).
 If you want to tell Python where to look for packages and modules,
 see [Python Handler: Finding modules](https://mkdocstrings.github.io/python/usage/#finding-modules).
+
--- a/src/mkdocstrings/plugin.py
+++ b/src/mkdocstrings/plugin.py
@@ -62,6 +62,7 @@ class MkdocstringsPlugin(BasePlugin):
         ("default_handler", MkType(str, default="python")),
         ("custom_templates", MkType(str, default=None)),
         ("enable_inventory", MkType(bool, default=None)),
+        ("enabled", MkType(bool, default=True)),
     )
     """
     The configuration options of `mkdocstrings`, written in `mkdocs.yml`.
@@ -72,6 +73,7 @@ class MkdocstringsPlugin(BasePlugin):
        Whenever a file changes in one of directories, the whole documentation is built again, and the browser refreshed.
        Deprecated in favor of the now built-in `watch` feature of MkDocs.
     - **`default_handler`**: The default handler to use. The value is the name of the handler module. Default is "python".
+    - **`enabled`**: Whether to enable the plugin. Default is true. If false, *mkdocstrings* will not collect or render anything.
     - **`handlers`**: Global configuration of handlers. You can set global configuration per handler, applied everywhere,
       but overridable in each "autodoc" instruction. Example:
 
@@ -128,6 +130,8 @@ class MkdocstringsPlugin(BasePlugin):
             *args: Additional arguments passed by MkDocs.
             **kwargs: Additional arguments passed by MkDocs.
         """
+        if not self.plugin_enabled:
+            return
         if self.config["watch"]:
             for element in self.config["watch"]:
                 log.debug(f"Adding directory '{element}' to watcher")
@@ -150,6 +154,9 @@ class MkdocstringsPlugin(BasePlugin):
         Returns:
             The modified config.
         """
+        if not self.plugin_enabled:
+            log.debug("Plugin is not enabled. Skipping.")
+            return config
         log.debug("Adding extension to the list")
 
         theme_name = None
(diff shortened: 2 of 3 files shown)


src/mkdocstrings/plugin.py
```diff
@@ -37,6 +37,8 @@
 """The name of the selection parameter in YAML configuration blocks."""
 RENDERING_OPTS_KEY: str = "rendering"
 """The name of the rendering parameter in YAML configuration blocks."""
+DISABLE_PLUGIN_ENV_VAR: str = "DISABLE_MKDOCSTRINGS" 
+"""The name of the environment variable that disables mkdocstrings."""
```
pawamoy, line 41: Rather than checking the environment in each hook, could we instead check it once here and use the value in the hooks?

```python
DISABLED: str =  os.getenv("DISABLE_MKDOCSTRINGS", "0").lower() in ("y", "yes", "true", "on", "1")
```

Then in the hooks:

```python
if DISABLED:
    return
```

I can't find a scenario where the environment variable would be updated in-between hooks. That would probably break everything anyway.

If assigning the value at import time is too early, we could maybe do it in the plugin init method instead.

WDYT?

oprypin (changes requested):
This should be part of the config, done like https://github.com/timvink/mkdocs-print-site-plugin/commit/315c2e018311965a821f343a17aaf5723926bb1d

pawamoy:
Thanks for the link @oprypin. What's the benefit of having the config value as well?

src/mkdocstrings/plugin.py
StefanBRas: I think that I don't know enough about the inner workings of mkdocs to say for sure how early it can be checked, so I just set it to be done as late as possible to be sure.

We could also log the debug message in the `on_startup` hook instead.

StefanBRas:
If we use the config, it's also possible to change it while hot reloading without restarting the process (but then we're back to forgetting to revert the change and accidentally committing it)

I think the test fails are unrelated to my change by the way, they're also failing on master for me.

pawamoy:
OK I checked the docs again and I see a clear benefit to having a configuration option: it allows users to enable/disable the plugin depending on *multiple* environment variables. It draws the path to a convention thanks to which you can disable all plugins at once, or just specific ones:

```yaml
plugins:
- plugin_a:
    enabled: !ENV [MKDOCS_ENABLE_PLUGINS, MKDOCS_ENABLE_PLUGINA, true]
- plugin_b: !ENV [MKDOCS_ENABLE_PLUGINS, MKDOCS_ENABLE_PLUGINB, true]
```

With this config, users can disable both plugins with `MKDOCS_ENABLE_PLUGINS=no` or disable just the plugin B with `MKDOCS_ENABLE_PLUGINB=no`.

This would allow even more grouping of plugins.

Of course all plugins should provide the same `enabled` or `disabled` options, as I don't think it's possible to say "not ENV_VAR".

So, @StephanBras, could you bring the change following what was done in the link @oprypin mentioned? Sorry for the late change of direction.

StefanBRas:
Sure, I'll do it over the weekend. Should it be an "enabled" option instead of "disabled" to be in line with others plugins then?

pawamoy:
Yes! Thanks a lot :)

src/mkdocstrings/plugin.py
pawamoy: That's irrelevant now that we want to add the config option anyway, resolving :)

StefanBRas: There's still the issue of multiple debug statements. I guess we could either:

- Don't log anything
- Log in every hook and don't care that we log multiple times
- Somehow check if we have logged the message already

I think the third one might be difficult, because when using `mkdocs serve` the user can change back and forth.

I think it's nice to get the debug message, just so it's very clear that the plugin is deliberately not being run.

But I have no strong feelings either way, I'll just make the PR without them and then they can be added in if needed later :)

pawamoy: Why not logging in the earliest hook and stop there?

StefanBRas: Oh yeah, I was thinking about only logging once per session. Logging in `on_config` should be good enough, right?

pawamoy: Yes, that's completely fine to have the debug log appear at each reload :slightly_smiling_face:

src/mkdocstrings/plugin.py
```diff
@@ -87,6 +88,8 @@ class MkdocstringsPlugin(BasePlugin):
               options:
                 selection_opt: 2
     ```
+
+    - **`enabled`**: Whether or not the plugin is enabled. Default is `True`. If false, all hooks exit early.
```
pawamoy, line 92:
```suggestion
    - **`enabled`**: Whether to enable the plugin. Default is true. If false, *mkdocstrings* will not collect or render anything.
```

Also please move the line up, after `default_handler` for example.

docs/usage.md
```diff
@@ -113,6 +113,7 @@ The above is equivalent to:
 - `handlers`: the handlers global configuration.
 - `enable_inventory`: whether to enable inventory file generation.
   See [Cross-references to other projects / inventories](#cross-references-to-other-projects-inventories)
+- `enabled` **(New in version 0.20)**: Whether or to enable the plugin. Defaults to `true`. Can be used to reduce build times when doing local deveopment. Especially useful when used with environment variables.
```
pawamoy, line 116:
```suggestion
- `enabled` **(New in version 0.20)**: Whether to enable the plugin. Defaults to `true`.
  Can be used to reduce build times when doing local development.
  Especially useful when used with environment variables (see example below).
```

docs/usage.md
```diff
@@ -121,6 +122,7 @@ The above is equivalent to:
     ```yaml title="mkdocs.yml"
     plugins:
     - mkdocstrings:
+        enabled: !ENV [ENABLE_MKDOCS, True]
```
pawamoy, line 125:
```suggestion
        enabled: !ENV [ENABLE_MKDOCSTRINGS, true]
```

pawamoy (changes requested):
Thanks! A few doc-related changes :slightly_smiling_face:

pawamoy (approved):
I've added a test :slightly_smiling_face:

Issue #502, opened 2023-01-04
If two different handlers are used in one document, the jinja env is not updated


jdpatt:
Found another issue/question with what I was trying to do with a custom handler.

If one document has two identifiers that use different handlers; only the first one's env is updated.  My templates error out because the default mkdocstring filters are missing (`highlight`, `heading` or `convert_markdown`) don't exist in the second handler yet and instead I only have the built-in jinja filters and [the `any` filter that is added when the `env` is initialized](https://github.com/mkdocstrings/mkdocstrings/blob/master/src/mkdocstrings/handlers/base.py#L111). 

```yaml
::: mkdocstrings_handlers.custom_handler.other_module.AccessTypes
    handler: python

::: demo.custom_identifier.stuff
    handler: custom_handler
```

It looks like [these few lines are the offenders](https://github.com/mkdocstrings/mkdocstrings/blob/master/src/mkdocstrings/extension.py#L200) and I'm curious if there is a drawback to updating it every time, updating the env on init or make the _updated_env a part of the handler and not the extension?  Commenting out `self._updated_env = True` is enough to resolve this issue.

```python
        if not self._updated_env:
            log.debug("Updating renderer's env")
            handler._update_env(self.md, self._config)  # noqa: WPS437 (protected member OK)
            self._updated_env = True
```

pawamoy:
Hey, nice catch! Indeed it does make sense to hold that `_updated_env` boolean at the processor level. Instead, each handler should have its own (or none). Do you want to open a PR?

jdpatt:
Yup, I can do that!

jdpatt:
Introduced in #201

PR #507, opened 2023-01-06
fix:  handle multiple handler's envs on same the doc

jdpatt:
Didn't want to take the full performance hit of doing it every time, so it now tracks if this autodoc instance has seen this handler before and if not; updates the handler's env.

Additionally, add a doc string to `_update_env` just a breadcrumb.

References: #201
Fixes: #502

---
I had issues with the instructions in `contributing.md` and had to manually install pdm on both windows/osx.  I got format working but check_types and test failed in locations not associated with this change.

### DIFF
--- a/src/mkdocstrings/extension.py
+++ b/src/mkdocstrings/extension.py
@@ -73,7 +73,7 @@ class AutoDocProcessor(BlockProcessor):
         self._config = config
         self._handlers = handlers
         self._autorefs = autorefs
-        self._updated_env = False
+        self._updated_envs: set = set()
 
     def test(self, parent: Element, block: str) -> bool:
         """Match our autodoc instructions.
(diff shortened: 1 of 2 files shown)


src/mkdocstrings/extension.py
```diff
@@ -78,7 +78,7 @@ def __init__(
         self._config = config
         self._handlers = handlers
         self._autorefs = autorefs
-        self._updated_env = False
+        self._updated_envs: list = []
```
pawamoy, line 81: I believe a `set` would be more appropriate.

pawamoy (changes requested):
Thanks a lot!

pawamoy:
And yeah CI breaks from time to time, I'll handle it.

src/mkdocstrings/extension.py
jdpatt: Updated.

pawamoy:
CI is fixed on master, you can rebase on it :slightly_smiling_face:

pawamoy (approved):
Thanks!

Issue #501, opened 2022-12-30
Handling non-css files on render


jdpatt:
Like everyone great tool that I use to document my python code.  I've been playing around with a custom handler based off your [template handler](https://github.com/mkdocstrings/handler-template) to document some internal file formats/code that are not python and have a few questions.  

I am using another tool to generate some image diagrams.  Doing that in the `render()` method makes mkdocs serve go into an infinite loop after you click a page since it sees a file change in the `docs_dir`, regenerates, sees lots of image files change again then loops.  I can do a few things but are looking to your suggestion on the best or if I have missed an easier path.  

### Use `mkdocs serve --dirtyreload` **[Current Solution]**
Not great since an update to that file doesn't necessarily trigger updating these diagrams since the are different files. So while I'm not infinitely looping, I'm not rendering the latest and greatest when serving.  Build works as expected.

### Use a similar recipe to [automatic code pages](https://mkdocstrings.github.io/recipes/#automatic-code-reference-pages) with `mkdocs-gen-files`.
Since this doesn't know what identifiers will be collected, I have to generate a lot of images that may never be used. I also have no guarantees when rendering that the file actually exists since I don't know the `site_dir` at the handler scope. 

### `mkdocstrings` change to expose mkdocs config to handler

Currently, it looks like the configuration is stripped down before passing to the handler.  A potential solution to give me access to the mkdocs configuration object and ultimate configurability by passing that to the handler?  This could allow `collect()`, `render()`, and `teardown()` have some knowledge about the current mkdocs configuration.  Looking at the python and crystal handlers, neither use `site_name`.

[plugin.py:168](https://github.com/mkdocstrings/mkdocstrings/blob/d965ccc39249971c897fba6a475b8f974e2de866/src/mkdocstrings/plugin.py#L168)
```python
        extension_config = {
            "site_name": config["site_name"],  # Change this to "mkdocs_config": config,
            "config_file_path": config["config_file_path"],
            "theme_name": theme_name,
            "mdx": config["markdown_extensions"],
            "mdx_configs": config["mdx_configs"],
            "mkdocstrings": self.config,
            # "mkdocs_config": config, #  Or add as a new item to not break existing code.
        }
        self._handlers = Handlers(extension_config)
```

### `mkdocstrings` change to add a `extra_files` method during the `on_env` event.

[...]

PR #509, opened 2023-01-11
feat: expose the full config to handlers

jdpatt:
Instead of just passing only the handler_config on initialization pass the full dictionary of `extension_config` directly to the handler.  Refactors items that were pulled from the full configuration to point directly to that object.

References: #501

### DIFF
--- a/docs/handlers/overview.md
+++ b/docs/handlers/overview.md
@@ -180,8 +180,8 @@ These arguments are all passed as keyword arguments, so you can ignore them
 by adding `**kwargs` or similar to your signature. You can also accept
 additional parameters: the handler's global-only options and/or the root
 config options. This gives flexibility and access to the mkdocs config, mkdocstring
-config and etc.. You should never modify the root config but can be used to get
-information about the mkdocs instance such as where the current `site_dir` lives.
+config etc.. You should never modify the root config but can use it to get
+information about the MkDocs instance such as where the current `site_dir` lives.
 See the [Mkdocs Configuration](https://www.mkdocs.org/user-guide/configuration/) for
 more info about what is accessible from it.
 
--- a/docs/handlers/overview.md
+++ b/docs/handlers/overview.md
@@ -134,7 +134,7 @@ Since version 0.14, you can create and use custom handlers
 thanks to namespace packages. For more information about namespace packages,
 [see their documentation](https://packaging.python.org/guides/packaging-namespace-packages/).
 
-TIP: **TL;DR - Project template for handlers.**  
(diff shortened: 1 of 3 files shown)


jdpatt:
>I'll have to think a bit more about the change, as it could allow us to clean up a lot of the underlying code.

@pawamoy interested to see where else you would see/want refactoring from having access to the full configuration object.  Keeping the same interfaces/backwards compatibility limited a few other places that I saw.  Namely in the `get_handler` call.  Would still need `options` passed to collect and render since the global options may have been overridden locally.

pawamoy:
Hey sorry, I forgot about this PR. Just so you know, it is fine to ping me if I don't answer :slightly_smiling_face: 
I'll try to review this shortly.

jdpatt:
> Hey sorry, I forgot about this PR. Just so you know, it is fine to ping me if I don't answer 🙂 I'll try to review this shortly.

No worries, we are all busy!

jdpatt:
@pawamoy , I'll ping you this time.  Did you get a chance to review?

docs/handlers/overview.md
```diff
@@ -134,7 +134,7 @@ Since version 0.14, you can create and use custom handlers
 thanks to namespace packages. For more information about namespace packages,
 [see their documentation](https://packaging.python.org/guides/packaging-namespace-packages/).
 
-TIP: **TL;DR - Project template for handlers.**  
+TIP: **TL;DR - Project template for handlers.**
```
pawamoy, line 137:
```suggestion
TIP: **TL;DR - Project template for handlers.**  
```

docs/handlers/overview.md
```diff
@@ -178,8 +178,12 @@ This function takes the following parameters:
 
 These arguments are all passed as keyword arguments, so you can ignore them
 by adding `**kwargs` or similar to your signature. You can also accept
-additional parameters: the handler's global-only options will be passed
-to this function when instantiating your handler.
+additional parameters: the handler's global-only options and/or the root
+config options. This gives flexibility and access to the mkdocs config, mkdocstring
+config and etc.. You should never modify the root config but can be used to get
+information about the mkdocs instance such as where the current `site_dir` lives.
```
pawamoy, line 184:
```suggestion
config etc.. You should never modify the root config but can use it to get
information about the MkDocs instance such as where the current `site_dir` lives.
```

pawamoy (approved):
OK lets start with that and I'll see later if there's a way to push it even further. Thanks!

jdpatt:
Sounds great, thanks!

Issue #510, opened 2023-01-16
intersphinx linking does not work with std:label entries but works with py:method ones


ssbarnea:
**Describe the bug**

Apparently any entries of type `std:label` from `objects.inv` are purely ignored/not-recognize, making impossible lo link to documentation.

**To Reproduce**

Lets analyze a practical example, https://docs.ansible.com/ansible/latest/objects.inv which can be converted to text for debugging via `sphobjinv convert plain objects.inv objects.txt`.

I will be listing some of the txt output here:

```
Ansible.get_bin_path py:method 1 api/index.html#$ -
developing_galaxy std:label -1 galaxy/dev_guide.html#developing-galaxy Galaxy Developer Guide
```

- Using `[Ansible.get_bin_path][]` works as expected.
- Using `[developing_galaxy][]` does not work and produces, `Could not find cross-reference target '[developing_galaxy]'`

It seems that `std:label` entries are not recognized but `py:method` ones are.

**Expected behavior**

Be able to link any entries from `objects.inv`.

**Information (please complete the following information):**
```
mkdocstrings==0.19.1
mkdocstrings-python==0.8.3
```

**Additional context**

ssbarnea:
In fact it seems that be a little bit different, as I seen also simple entries not being recognized but others being. 

Lets analyze a practical example, https://docs.ansible.com/ansible/latest/objects.inv which can be converted to text for debugging via `sphobjinv convert plain objects.inv objects.txt`.

I will be listing some of the txt output here:

```
Ansible.get_bin_path py:method 1 api/index.html#$ -
developing_galaxy std:label -1 galaxy/dev_guide.html#developing-galaxy Galaxy Developer Guide
```

## Outcomes

- Using `[Ansible.get_bin_path][]` works as expected.
- Using `[developing_galaxy][]` does not work and produces, `Could not find cross-reference target '[developing_galaxy]'`

It seems that `std:label` entries are not recognized but `py:method` ones are.

pawamoy:
Hello, thanks for the investigation and report! Since we claim to support Sphinx inventories, I guess this can be considered a bug indeed.

We can see here that we explicitly filter on `py` domain:

https://github.com/mkdocstrings/python/blob/4ce0fc824fc36cadcf2b70cc72b627bdaa5abd65/src/mkdocstrings_handlers/python/handler.py#L189

...probably because this is done in the Python handler, which has nothing to do with the `std` domain :thinking: 

[...]

PR #511, opened 2023-01-17
Make _load_inventory accept lists as arguments

ssbarnea:
This change should enable us to pass any kind of configuration arguments to the inventory
and unblock https://github.com/mkdocstrings/python/pull/49 which was not able to load
sequences as PYYAML loads them as lists, which are mutable and not allowed by pycache.

As this method is not expected to modify any of the received data, there is no problem to
convert its non-mutable list arguments into tuples using a small wrapper.

Partial: https://github.com/mkdocstrings/mkdocstrings/issues/510

### DIFF
--- a/src/mkdocstrings/plugin.py
+++ b/src/mkdocstrings/plugin.py
@@ -42,6 +42,18 @@ InventoryImportType = List[Tuple[str, Mapping[str, Any]]]
 InventoryLoaderType = Callable[..., Iterable[Tuple[str, str]]]
 
 
+def list_to_tuple(function: Callable[..., Any]) -> Callable[..., Any]:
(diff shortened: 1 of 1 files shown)


pawamoy (approved):
Ha, didn't think of decorating the function! I like it, thanks :slightly_smiling_face: 
Quality checks will probably fail though, variables must have names longer than 1 character, also it will complain about missing parameters in the docstring, so you can just remove the docstring and make the decorator private.

Maybe just add a comment above the decorators (or next to `@list_to_tuple` if it's short enough) to explain why we need to convert lists to tuples, so we don't forget.

pawamoy:
Thanks a lot!

PR #520, opened 2023-01-21
fix: correct type signature for BaseCollector.collect

ssbarnea:
This is needed by python library because `collect()` needs to pass that argument to a method that declares it as MutableMapping.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,7 +114,7 @@ tests = [
     "sphinx",
 ]
 typing = [
-    "mypy>=0.910",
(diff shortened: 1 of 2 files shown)


pawamoy:
@ssbarnea should I wait for you to mark this PR as ready for review?

ssbarnea:
@pawamoy No need to wait, if you find the draft PR as good enough, feel free to convert it and merge. Thanks for the ping, I forgot, too many changes in too many repos....

pawamoy (approved):

PR #531, opened 2023-02-16
Update language extras install examples with pip

lognaturel:
I'm just getting started. I use Python fairly frequently but am not a tooling expert and don't know a lot about package naming conventions. Here was my journey:
- First I copied `pip install mkdocstrings[crystal,python]` exactly and that didn't work.
- Then I figured maybe `[crystal,python]` was intended as some kind of disjunction notation so I tried `mkdocstringpython` and that didn't work.
- Then I followed the link to https://mkdocstrings.github.io/handlers/overview/ and didn't see anything about pip installation.
- Then I clicked on the experimental link and made my way to https://mkdocstrings.github.io/python/
- Then I saw `mkdocstrings-python` in the PEP 621 declaration and tried `pip install mkdocstrings-python`

That feels like a lot of unnecessary friction! I think the clearest is to have the two examples one after the other as I have proposed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ pip install mkdocstrings
 You can install support for specific languages using extras, for example:
 
 ```bash
-pip install mkdocstrings[crystal,python]
+pip install 'mkdocstrings[crystal,python]'
 ```
 
 See the [available language handlers](https://mkdocstrings.github.io/handlers/overview/).


lognaturel:
The test failure looks spurious. I can't quickly tell whether the quality issue is related to my change, sorry.

pawamoy:
Hello, thanks for the PR and sorry for the trouble!

I'd prefer that we simply quote the argument. As you guessed, in some shells the brackets `[]` are interpreted. If you quote the argument, they stop being interpreted. Some shells don't need the quotes.

```
pip install 'mkdocstrings[crystal,python]'
```

Could you update the PR like so?
Also yeah, don't pay attention to these CI failures, they're not related to your change.

lognaturel:
Oh, wow, TIL about [optional dependencies](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies). I've never run into that before!

pawamoy:
Thanks for the links!

For some reason, brackets are not interpreted by my Zsh shell, so I'm able to type these commands without quotes. I guess it depends on how Zsh is configured :shrug:

pawamoy (approved):

lognaturel:
Merci! 😊

pawamoy:
Merci à toi :rocket:

PR #569, opened 2023-05-12
feat: Allow extensions to add templates

pawamoy:
An extension here is simply a Python package that defines an entry-point for a specific handler.

For example, an extension can add templates to the Python handler thanks to this entry-point:

```toml
[project.entry-points."mkdocstrings.python.templates"]
extension-name = "extension_package:get_templates_path"
```

This entry-point assumes that the extension provides a `get_templates_path` function directly under the `extension_package` package. This function doesn't accept any argument and returns the path to a directory containing templates. The directory must contain one subfolder for each supported theme, for example:

```
templates/
    material/
    readthedocs/
    mkdocs/
```

mkdocstrings will add the folders corresponding to the user-selected theme, and to the handler's defined fallback theme, as usual.

The names of the extension templates must not overlap with the handler's original templates.

The extension is then responsible, in collaboration with its target handler, for mutating the collected data in order to instruct the handler to use one of the extension template when rendering particular objects.

For example, the Python handler will look for a `template` attribute on objects, and use it to render the object. This `template` attribute will be set by Griffe extensions (Griffe is the tool used by the Python handler to collect data).
See related PR: https://github.com/mkdocstrings/python/pull/70

----

### TODO

- [x] docs
- [x] tests

### DIFF
--- a/docs/usage/handlers.md
+++ b/docs/usage/handlers.md
@@ -192,7 +192,7 @@ is written for inspiration.
 ### Templates
 
 Your handler's implementation should normally be backed by templates, which go
-to the directory `mkdocstrings_handlers/custom_handler/templates/some_theme`.
+to the directory `mkdocstrings_handlers/custom_handler/templates/some_theme`
 (`custom_handler` here should be replaced with the actual name of your handler,
 and `some_theme` should be the name of an actual MkDocs theme that you support,
 e.g. `material`).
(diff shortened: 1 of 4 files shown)


oprypin:
Thanks.
No comments from me 🙂
Generally you don't need to wait for my review. Or let's say... if I haven't commented on something in the first 2 days, I probably won't.

PR #575, opened 2023-06-08
feat: allow ::: pattern in indented blocks

percevalw:
Hi,

This PR extends the autodoc syntax to allow inserting autodocs in indented blocks.
This allows, for instance, to write the following (collapsible autdoc):

```md
??? note "Reference"

    ::: my_project.my_file.my_class
        options:
          heading_level: 3
```

which was not possible before.

Let me know what you think !

### DIFF
--- a/src/mkdocstrings/extension.py
+++ b/src/mkdocstrings/extension.py
@@ -59,7 +59,7 @@ class AutoDocProcessor(BlockProcessor):
     a matched block.
     """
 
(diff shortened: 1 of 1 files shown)


pawamoy:
I think it is already working. At least I just tried it locally and the reference is correctly injected within the "Reference" collapsible admonition.

percevalw:
Indeed I've reinstalled and it works. I don't know what caused the issue with admonitions in the first place. Sorry for the noise !

pawamoy:
No worries! Thanks for closing :slightly_smiling_face:

PR #590, opened 2023-07-17
refactor: Remove deprecated parts

pawamoy:
Removed:

- Base collector class
- Base renderer class
- Watch feature
- Selection and rendering keys support (YAML options)
- `mkdocstrings.handlers` namespace (when importing handlers and finding templates)

### DIFF
--- a/config/mypy.ini
+++ b/config/mypy.ini
@@ -3,5 +3,3 @@ ignore_missing_imports = true
 exclude = tests/fixtures/
 warn_unused_ignores = true
 show_error_codes = true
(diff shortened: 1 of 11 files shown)


pawamoy:
The `invalidate_caches` error in CI is already fixed in editables, just waiting a release.

pawamoy:
PyYAML issue: https://github.com/yaml/pyyaml/issues/724. It does not provide wheels for Python 3.12 yet, so it's being built from sources and errors with Cython 3 which just got released.

PR #593, opened 2023-07-26
Add note about newline handling

ssweber:
I added a note about new-line handling. Maybe its a bug that I should be reporting elsewhere?

```python
class ExampleClass:
    """This class is broken because of \n

    Example:
    ```python
    msg = "There should be a newline char at the end of this example\n"
    print(msg)
    ```
    """
    pass
```

produces this:

> ![image](https://github.com/mkdocstrings/mkdocstrings/assets/57631333/3ae8d7ca-5308-469d-a2b8-cc3e3bc0e517)

Escaping the "\n" with an additional "\" produces:

> ![image](https://github.com/mkdocstrings/mkdocstrings/assets/57631333/49f957a6-1c6c-490d-b4b2-a7f827d39044)

my mkdocs.yaml:
```mkdocs
markdown_extensions:
    - admonition
    - codehilite
    - pymdownx.superfences

plugins:
- search
- mkdocstrings:
    handlers:
      python:
        options:
          docstring_style: "google"

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -21,6 +21,8 @@ markdown_extensions:
 - pymdownx.superfences
 ```
 
+Note: If a code-block has a newline, "\n", you must escape it with an additional "\\", eg ```msg = "This line has a newline char at the end \\n"```
+
 ## Footnotes are duplicated or overridden
 
 Before version 0.14, footnotes could be duplicated over a page.
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -21,8 +21,8 @@ markdown_extensions:
 - pymdownx.superfences
 ```
 
-For code blocks in docstrings, make sure to escape newlines: `\n` -> `\\n`, or prefix
-the entire docstring with 'r' to make it a raw-docstring: <code>r"""</code>.
+For code blocks in docstrings, make sure to escape newlines (`\n` -> `\\n`),
+or prefix the entire docstring with 'r' to make it a raw-docstring: `r"""`.
 Indeed, docstrings are still strings and therefore subject to how Python parses strings.
 
 ## Footnotes are duplicated or overridden
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -21,7 +21,9 @@ markdown_extensions:
 - pymdownx.superfences
 ```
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ssweber** [comment]
I’ll check when I have a second if it still breaks if you do a raw docstring, eg r”””

For example pydoctor says:
>Since docstrings are Python strings, escape sequences such as \n will be parsed as if the corresponding character—for example a newline—occurred at the position of the escape sequence in the source code.

>To have the text \n in a docstring at runtime and in the generated documentation, you either have escape it twice in the source: \\n or use the r prefix for a raw string literal. The following example shows the raw string approach:

>def iter_lines(stream):
    r"""Iterate through the lines in the given text stream,
    with newline characters (\n) removed.
    """
    for line in stream:
        yield line.rstrip('\n')

File: docs/troubleshooting.md [hunk 1]
```diff
@@ -21,6 +21,8 @@ markdown_extensions:
 - pymdownx.superfences
 ```
 
+Note: If a code-block has a newline, "\n", you must escape it with an additional "\\", eg ```msg = "This line has a newline char at the end \\n"```
```
pawamoy, line 24:
```suggestion
For code blocks in docstrings, make sure to escape newlines: `\n` -> `\\n`.
Indeed, docstrings are still strings and therefore subject to how Python parses strings.

**Review by pawamoy — COMMENTED** [review]
Alternative suggestion, WDYT?

**Comment by ssweber** [comment]
I like your edit. I tested <code>r"""</code>, and it works as well, so I added that to the comment

File: docs/troubleshooting.md [hunk 2]
```diff
@@ -21,6 +21,10 @@ markdown_extensions:
 - pymdownx.superfences
 ```
 
+For code blocks in docstrings, make sure to escape newlines: `\n` -> `\\n`, or prefix
+the entire docstring with 'r' to make it a raw-docstring: <code>r"""</code>.
```
pawamoy, line 25:
```suggestion
For code blocks in docstrings, make sure to escape newlines (`\n` -> `\\n`),
or prefix the entire docstring with 'r' to make it a raw-docstring: `r"""`.
```

pawamoy (approved):
Thanks!

ssweber:
thank you!

Issue #609, opened 2023-09-17
Duplicated headings in nested blocks


percevalw:
**Describe the bug**

Nesting a docstring in another block (e.g., tab or admonition) produces duplicated headings.

<h5 a><strong><code>tests/fixtures/headings.py</code></strong></h5>

```python
"""
Foo
===

### Bar

###### Baz
"""
```

<h5 a><strong><code>index.md</code></strong></h5>

```markdown
Top.[^aaa]

=== "Tab A"

    ::: tests.fixtures.headings
```

In this case, the "Foo" header will appear twice in the produced html, both at the right position and after the inserted docstring :

```html
<p>Top.[^aaa]</p>
  <div class="tabbed-set tabbed-alternate" data-tabs="1:1"><input checked="checked" id="__tabbed_1_1" name="__tabbed_1" type="radio" /><div class="tabbed-labels"><label for="__tabbed_1_1">Tab A</label></div>
  <div class="tabbed-content">
  <div class="tabbed-block">
  <div class="mkdocstrings">

  <div class="doc doc-object doc-module">



  <a id="tests.fixtures.headings"></a>
    <div class="doc doc-contents first">

        <h2 id="tests.fixtures.headings--foo">Foo</h2>
  <h4 id="tests.fixtures.headings--bar">Bar</h4>
  <h6 id="tests.fixtures.headings--baz">Baz</h6>

    </div>

  </div><h2 data-role="module" id="tests.fixtures.headings"></h2>
  <h2 id="tests.fixtures.headings--foo">Foo</h2>
  <h4 id="tests.fixtures.headings--bar">Bar</h4>
  <h6 id="tests.fixtures.headings--baz">Baz</h6>
  </div>
  </div>
  </div>
  </div>
```

**Expected behavior**

Headings from docstrings should not be duplicated.

PR #610, opened 2023-09-17
Remove duplicated headings for docstrings nested in tabs/admonitions

percevalw:
Hi,

As mentioned in #609, nesting a docstring in another block (e.g., tab or admonition) produces duplicated headings.
At the moment, the extension postprocessor only checks inside the immediate root children when removing duplicated heading.

Some markdown extensions move the produced docstring in a nested html node : we only need to search deeply when removing duplicatas.

### DIFF
--- a/src/mkdocstrings/extension.py
+++ b/src/mkdocstrings/extension.py
@@ -231,17 +231,26 @@ class AutoDocProcessor(BlockProcessor):
 
 class _PostProcessor(Treeprocessor):
     def run(self, root: Element) -> None:
+        self.remove_duplicated_headings_from_parent(root)
+
+    def remove_duplicated_headings_from_parent(self, parent: Element) -> bool:
         carry_text = ""
-        for el in reversed(root):  # Reversed mainly for the ability to mutate during iteration.
+        found = False
+        for el in reversed(parent):
             if el.tag == "div" and el.get("class") == "mkdocstrings":
                 # Delete the duplicated headings along with their container, but keep the text (i.e. the actual HTML).
                 carry_text = (el.text or "") + carry_text
-                root.remove(el)
+                parent.remove(el)
+                found = True
             elif carry_text:
                 el.tail = (el.tail or "") + carry_text
                 carry_text = ""
+            elif self.remove_duplicated_headings_from_parent(el):
+                found = True
+                break
         if carry_text:
-            root.text = (root.text or "") + carry_text
+            parent.text = (parent.text or "") + carry_text
+        return found
 
 
 class MkdocstringsExtension(Extension):
--- a/src/mkdocstrings/extension.py
+++ b/src/mkdocstrings/extension.py
@@ -231,12 +231,12 @@ class AutoDocProcessor(BlockProcessor):
 
 class _PostProcessor(Treeprocessor):
     def run(self, root: Element) -> None:
-        self.remove_duplicated_headings_from_parent(root)
+        self._remove_duplicated_headings(root)
 
-    def remove_duplicated_headings_from_parent(self, parent: Element) -> bool:
+    def _remove_duplicated_headings(self, parent: Element) -> bool:
         carry_text = ""
         found = False
-        for el in reversed(parent):
+        for el in reversed(parent):  # Reversed mainly for the ability to mutate during iteration.
             if el.tag == "div" and el.get("class") == "mkdocstrings":
                 # Delete the duplicated headings along with their container, but keep the text (i.e. the actual HTML).
                 carry_text = (el.text or "") + carry_text
@@ -245,7 +245,7 @@ class _PostProcessor(Treeprocessor):
             elif carry_text:
                 el.tail = (el.tail or "") + carry_text
                 carry_text = ""
-            elif self.remove_duplicated_headings_from_parent(el):
+            elif self._remove_duplicated_headings(el):
                 found = True
                 break
         if carry_text:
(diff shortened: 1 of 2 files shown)


pawamoy:
Amazing! I can confirm that both the issue exists and the fix works!

src/mkdocstrings/extension.py
```diff
@@ -231,17 +231,26 @@ def _process_block(
 
 class _PostProcessor(Treeprocessor):
     def run(self, root: Element) -> None:
+        self.remove_duplicated_headings_from_parent(root)
+
+    def remove_duplicated_headings_from_parent(self, parent: Element) -> bool:
```
pawamoy, line 236: Nitpick: I find "from parent" confusing since it iterates downwards. Let's just call it `_remove_duplicated_headings`.

```suggestion
    def _remove_duplicated_headings(self, parent: Element) -> bool:
```

src/mkdocstrings/extension.py
```diff
@@ -231,17 +231,26 @@ def _process_block(
...
         carry_text = ""
-        for el in reversed(root):  # Reversed mainly for the ability to mutate during iteration.
+        found = False
+        for el in reversed(parent):
```
pawamoy, line 239: Lets keep the previous comment:

```suggestion
        for el in reversed(parent):  # Reversed mainly for the ability to mutate during iteration.
```

src/mkdocstrings/extension.py
```diff
@@ -231,17 +231,26 @@ def _process_block(
...
             if el.tag == "div" and el.get("class") == "mkdocstrings":
                 # Delete the duplicated headings along with their container, but keep the text (i.e. the actual HTML).
                 carry_text = (el.text or "") + carry_text
-                root.remove(el)
+                parent.remove(el)
+                found = True
             elif carry_text:
                 el.tail = (el.tail or "") + carry_text
                 carry_text = ""
+            elif self.remove_duplicated_headings_from_parent(el):
```
pawamoy, line 248:
```suggestion
            elif self._remove_duplicated_headings(el):
```

tests/test_extension.py
```diff
@@ -150,3 +150,23 @@ def test_use_options_yaml_key(ext_markdown: Markdown) -> None:
     """Check that using the 'options' YAML key works as expected."""
     assert "h1" in ext_markdown.convert("::: tests.fixtures.headings\n    options:\n      heading_level: 1")
     assert "h1" not in ext_markdown.convert("::: tests.fixtures.headings\n    options:\n      heading_level: 2")
+
+
+@pytest.mark.parametrize("ext_markdown", [{"markdown_extensions": [{"pymdownx.tabbed": {"alternate_style": True}}]}], indirect=["ext_markdown"])
+def test_tab_headings(ext_markdown: Markdown) -> None:
```
pawamoy, line 156:
```suggestion
def test_removing_duplicated_headings(ext_markdown: Markdown) -> None:
```

tests/test_extension.py
```diff
@@ -150,3 +150,23 @@ def test_use_options_yaml_key(ext_markdown: Markdown) -> None:
...
+    """Assert footnotes don't get added to subsequent docstrings."""
```
pawamoy, line 157:
```suggestion
    """Assert duplicated headings are removed from the output."""
```

tests/test_extension.py
```diff
@@ -150,3 +150,23 @@ def test_use_options_yaml_key(ext_markdown: Markdown) -> None:
...
+    output = ext_markdown.convert(
+        dedent(
+            """
+            Top.[^aaa]
+
```
pawamoy, line 162: I don't think the footnote is useful to the test here.

```suggestion
```

pawamoy (changes requested):
Thanks a lot! I had noticed the issue but didn't get time to report it. You reported it, and fixed it, that's awesome :heart_eyes: 

Just a few nitpicks!

pawamoy (commented):
OK perfect, thanks so much again!

src/mkdocstrings/extension.py
```diff
@@ -231,17 +231,26 @@ def _process_block(
...
             if el.tag == "div" and el.get("class") == "mkdocstrings":
                 # Delete the duplicated headings along with their container, but keep the text (i.e. the actual HTML).
                 carry_text = (el.text or "") + carry_text
-                root.remove(el)
+                parent.remove(el)
+                found = True
             elif carry_text:
                 el.tail = (el.tail or "") + carry_text
                 carry_text = ""
+            elif self._remove_duplicated_headings(el):
```
oprypin, line 248: This has a wrong behavior: "If text was carried, don't stepping into the current element". There is no connection between the two, so it shouldn't be an `else`.

src/mkdocstrings/extension.py
```diff
@@ -231,17 +231,26 @@ def _process_block(
...
+                found = True
+                break
```
oprypin, line 250: This has a wrong behavior: "If duplicated headings were found in a child element, stop scanning the rest of the document upwards". Why?

pawamoy: I guess there's no specific "why" and the assumption is that duplicated headings appear in a single child element. Do you recommend to try and remove them from the whole subtree?

oprypin: Results from the child element are being propagated to the parent loop and it gets interrupted.

PR #629, opened 2023-10-26
Drop support for MkDocs < 1.4, modernize usages

oprypin:


src/mkdocstrings/plugin.py
```diff
@@ -63,38 +62,13 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
...
-
-    - **`handlers`**: Global configuration of handlers. You can set global configuration per handler, applied everywhere,
-      but overridable in each "autodoc" instruction. Example:
-    - **`default_handler`**: The default handler to use. The value is the name of the handler module. Default is "python".
-    - **`custom_templates`**: Location of custom templates to use when rendering API objects. Value should be the path of 
-      a directory relative to the MkDocs configuration file.
-    - **`enable_inventory`**: Whether to enable object inventory creation.
-    - **`enabled`**: Whether to enable the plugin. Default is true. If false, *mkdocstrings* will not collect or render anything.
+    Global configuration of handlers. You can set global configuration per handler, applied everywhere,
+    but overridable in each "autodoc" instruction. Example:
```
pawamoy, line 71:
```suggestion
    Global configuration of handlers.

    You can set global configuration per handler, applied everywhere,
    but overridable in each "autodoc" instruction. Example:
```

src/mkdocstrings/plugin.py
```diff
@@ -110,6 +84,30 @@ class MkdocstringsPlugin(BasePlugin):
     ```
     """
 
+    default_handler = opt.Type(str, default="python")
+    """The default handler to use. The value is the name of the handler module. Default is "python"."""
+    custom_templates = opt.Optional(opt.Dir(exists=True)),
+    """Location of custom templates to use when rendering API objects. Value should be the path of
+    a directory relative to the MkDocs configuration file."""
```
pawamoy, line 91:
```suggestion
    """Location of custom templates to use when rendering API objects.
    
    Value should be the path of a directory relative to the MkDocs configuration file.
    """
```

pawamoy (approved):
Awesome, thanks!

Just two suggestions for docstrings, to separate the summary from the body when its too long for a single line.

PR #631, opened 2023-11-12
Fix tests, fix/ignore typing

pawamoy:


oprypin:
Oof, I caused these complications with types 🥲

Each of those ignored places could be better written with a cast, but on the other hand casts kinda suck because they are a runtime operation and don't work with `if TYPE_CHECKING`

pawamoy:
No worries, CI was failing for other reasons so we both missed the tests failure. Also yeah, I really dislike using `cast`. I used it once here because it was easy enough. Maybe I should just remove it and add a type ignore comment for consistency.

oprypin:
No I mean I made the change to type annotations to be *more correct* but maybe they should've stayed less correct 😂

https://github.com/python/typeshed/pull/10963/files#diff-fdd5d1aae42754416614f2980c7ce6aeacb412d8614b5923858bf517a627f9f0

pawamoy:
Oh OK, I thought you were speaking about the PR that modernized the tests to use `MkDocsConfig`.

More correct is good!

oprypin (approved):

PR #632, opened 2023-11-12
feat: Cache downloaded inventories as local files

oprypin:


pyproject.toml
```diff
@@ -29,11 +29,13 @@ classifiers = [
     "Typing :: Typed",
 ]
 dependencies = [
+    "click>=7.0",
```
pawamoy, line 32: Can we avoid the dependency :confused: ?

src/mkdocstrings/_cache.py
```diff
@@ -0,0 +1,76 @@
...
+                        log.debug(f"Using cached '{path}' for '{url}'")
+                        return f.read()
+        except (OSError, ValueError) as e:
+            log.debug(f"{type(e).__name__}: {e}")
+
+    # Download and cache the file
+    log.debug(f"Downloading '{url}' to '{path}'")
+    content = download(url)
+    os.makedirs(directory, exist_ok=True)
+    with click.open_file(path, "wb", atomic=True) as f:
```
pawamoy, line 73: So if I follow correctly, this function gets executed in a thread pool, and that's why we want atomic file writes. But in what circumstances could we have multiple writes on the same path? When the user is for example running `mkdocs build` and `mkdocs serve` at the same time :thinking: ? Or if they listed twice or more the same inventory in `mkdocs.yml`? In the worst case, the cache gets corrupted, but cache corruption seems to be handled above, in which case the file is simply re-downloaded. All in all, is the added `click` dependency worth it?

pawamoy (commented):
Should I merge #631 first? Looks like you based your branch off of it.

pyproject.toml
oprypin: It is already a dependency so it doesn't matter. "mkdocstrings -> mkdocs -> click" will always be there

src/mkdocstrings/_cache.py
oprypin: > So if I follow correctly, this function gets executed in a thread pool, and that's why we want atomic file writes

No, this is if the write is interrupted in the middle and then you get a half-downloaded file cached permanently.

pyproject.toml
pawamoy: Hmmm, true, thanks for the reminder.

src/mkdocstrings/_cache.py
pawamoy: OK, my main concern was the "added" Click dependency, but it's actually not added, as you explained in the other comment, so let's make good use of it.

pawamoy (approved):

src/mkdocstrings/_cache.py
oprypin: Ha. Now that I tried to actually replace it with something simpler, I ran into a race condition quite easily :smile: 

https://github.com/oprypin/mkdocs-gallery/actions/runs/6928061416/job/18843101055

mkdocs-gallery runs multiple `get-deps` at the same time

pawamoy: My cross-platform, cross-language, dependency-free, dumb solution for file locking is to try and create directories as locks :rofl: Well I know it works on Linux because directory creation is atomic. I implement blocking (waiting) with a simple while loop + time sleep :shrug:

PR #649, opened 2024-01-27
Circumvent error with mkdocs link.

MrCurtis:
Hi,

Currently navigating to https://mkdocs.org/ causes a certification error.

mkdocs_error

Changing the link to https://www.mkdocs.org/ circumvents this.


pawamoy (approved):
Thanks, good catch!

PR #651, opened 2024-02-20
refactor: Backup anchors without hrefs, for compatibility with autorefs' Markdown anchors

pawamoy:
This will be useful for the soon-to-come Markdown anchors feature of mkdocs-autorefs.

Related to mkdocs-autorefs#39: https://github.com/mkdocstrings/autorefs/pull/39

### DIFF
--- a/src/mkdocstrings/handlers/rendering.py
+++ b/src/mkdocstrings/handlers/rendering.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import copy
 import re
 import textwrap
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Iterator
 
 from markdown.extensions import Extension
 from markdown.extensions.codehilite import CodeHiliteExtension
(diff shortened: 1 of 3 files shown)


pawamoy:
I first went with a custom iterator approach (yielding parent, index, element) to insert at the right position, but since it iterated **forward** it was causing infinite loops.

Instead, I used a recursive approach, and at each layer I iterate on elements **backwards** to be able to insert in the being-iterated elements. The other option was to iterate on a copy of the list of elements, but this would have been less efficient. `reversed` is possible each because `Element` implements both `__len__` and `__getitem__`.

src/mkdocstrings/handlers/rendering.py
```diff
@@ -146,19 +146,29 @@ def __init__(self, md: Markdown, id_prefix: str):
...
-        for el in root.iter():
-            id_attr = el.get("id")
-            if id_attr:
-                el.set("id", self.id_prefix + id_attr)
+        if self.id_prefix:
+            self._prefix_ids(root)
+
+    def _prefix_ids(self, root: Element) -> None:
+        index = -1
+        for el in reversed(root):
```
pawamoy, line 154: Could add a comment such as in the post processor here:

> Reversed mainly for the ability to mutate during iteration.

src/mkdocstrings/handlers/rendering.py
```diff
@@ -146,19 +146,29 @@ def __init__(self, md: Markdown, id_prefix: str):
...
+            index += 1
 
+            self._prefix_ids(el)
```
pawamoy, line 157: We descend early. Doing it after handling the element works too. No strong opinion.

src/mkdocstrings/handlers/rendering.py
```diff
@@ -146,19 +146,29 @@ def __init__(self, md: Markdown, id_prefix: str):
...
             href_attr = el.get("href")
+
+            if id_attr := el.get("id"):
+                if el.tag == "a" and not href_attr:
+                    new_el = copy.deepcopy(el)
```
pawamoy, line 162: `copy.copy` was not enough: `new_el.set` would mutate the original element.

src/mkdocstrings/handlers/rendering.py
```diff
@@ -146,19 +146,35 @@ def __init__(self, md: Markdown, id_prefix: str):
...
-            id_attr = el.get("id")
-            if id_attr:
-                el.set("id", self.id_prefix + id_attr)
+        if self.id_prefix:
+            self._prefix_ids(root)
 
+    def _prefix_ids(self, root: Element) -> None:
+        index = -1
+        for el in reversed(root):  # Reversed mainly for the ability to mutate during iteration.
+            index += 1
```
oprypin, line 155: The iteration is in reverse order but the indices are in forward order :thinking: so this will fail with multiple elements present

Please consider fetching and merging my fix commit https://github.com/oprypin/mkdocstrings/commit/735e2966db85779cdd459ec6ff6976218d2c0580

oprypin:
Looks awesome otherwise :)

src/mkdocstrings/handlers/rendering.py
pawamoy: Of the importance of code reviews... How did I miss that 🫠
Thanks!

pawamoy:
OK LGTM! Squashing incoming :slightly_smiling_face:

PR #696, opened 2024-10-03
All

ElementalWarrior:
main

### DIFF
--- a/docs/usage/index.md
+++ b/docs/usage/index.md
@@ -116,6 +116,7 @@ The above is equivalent to:
 - `enabled` **(New in version 0.20)**: Whether to enable the plugin. Defaults to `true`.
   Can be used to reduce build times when doing local development.
   Especially useful when used with environment variables (see example below).
+- `css_filename`: The css path and filename to create the stylesheet.
 
 !!! example
     ```yaml title="mkdocs.yml"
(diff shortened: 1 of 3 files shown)


ElementalWarrior:
Hello @pawamoy. Would you consider giving this a review?

pawamoy:
Hey again @ElementalWarrior, thanks for this PR too!

> the css filename doesn't pass the DOMPurify filtering

Can you elaborate? What's this filtering? Should we just change the default name, without providing a new option that requires maintenance? I don't expect anyone to rely on the current name anyway, so I think it would be safe to change it. What value would you give it if this PR was merged?

ElementalWarrior:
Backstage attempts to white list what resources can be loaded to prevent security issues. In this case they white list link src attribute to like specific Google domains, and a regex of `main.[a-fA-F0-9]{8}.min.css`.

I just ended up renaming the file to mkdocstrings.main.12345678.min.css to pass it's filtering.

pawamoy:
I see, thanks. Unfortunately, you'll have the same issue with other plugins that insert their own stylesheets, so it doesn't seem worth adding and maintaining a configuration option here just to alleviate the strictness/oversight of Backstage. I'd prefer we have a discussion with them first, if possible. Would you like to open a discussion on their repo, linking to this PR maybe?

Issue #707, opened 2024-11-01
feature: Download private inventory files


connesy:
### Is your feature request related to a problem? Please describe.
In my company, we have a project `Foo`, in which we build the documentation with `mkdocs` and `mkdocstrings`. The generated site and `objects.inv` is then pushed to a private GitHub repository to publish the documentation on a private site through GitHub Pages.

Now we are trying to build the documentation for another project `Bar`, which has references to objects in `Foo`, however, since the inventory file for `Foo` is placed in a private repo, it cannot be downloaded as part of `mkdocs build`.

### Describe the solution you'd like
It would be great to have some way to be able to either pass credentials and/or have them be read from e.g. environment variables or `.netrc` when downloading inventory files.

### Describe alternatives you've considered
Manually downloading the inventory file(s) and pointing to the local file path in `mkdocs.yml` when building the documentation. This becomes a real hassle when dealing with a team of developers having to manually download multiple inventory files and changing `mkdocs.yml` to point to their local file system.

We are really happy with `mkdocstrings`, this is just the missing piece for us :slightly_smiling_face:

pawamoy:
Hi @connesy, thanks for the feature request. So you want to be able to do this:

```
plugins:
- mkdocstrings:
    handlers:
      python:
        import:
        - https://${CREDENTIALS}@domain.com/etc/objects.inv
```

...where `CREDENTIALS` is an env var with a value like `username:password`. Do I understand correctly?

connesy:
Hi @pawamoy 

Yes, that would be a great solution! `CREDENTIALS` would have to be able to be either `username:password` or a single `api_token`, as is often used with e.g. GitHub.

pawamoy:
Yes you'd be able to put whatever you want in the environment variable and build your URL with it :slightly_smiling_face: I'll go with a naive solution first and not do any escaping/URL-encoding.

connesy:
@pawamoy I have previously tested adding a token directly to the URL, but `urllib` failed to parse the URL upon making the request. It took the `token@github.com` as being the host.

pawamoy:
Looks like you're right:

```pycon
>>> urlparse("https://username:password@example.com/path")
ParseResult(scheme='https', netloc='username:password@example.com', path='/path', params='', query='', fragment='')
```

We can probably parse the URL, extract the username and password or the token, then rebuild the URL and pass the auth credentials accordingly.

connesy:
[...]

PR #710, opened 2024-11-05
Add support for authentication in inventory file URLs

connesy:
Implement basic and bearer token authentication for downloading private inventory files in mkdocstrings. This allows users to provide credentials via environment variables in the URL. 

Update documentation to reflect these changes and add test cases to ensure functionality.

Closes https://github.com/mkdocstrings/mkdocstrings/issues/707.

### DIFF
--- a/docs/usage/index.md
+++ b/docs/usage/index.md
@@ -31,8 +31,8 @@ The YAML block is optional, and contains some configuration options:
   `default_handler` key, or `"python"`.
 - `options`: a dictionary of options passed to the handler's methods responsible both
   for collecting and rendering the documentation. These options can be defined
-  globally (in `mkdocs.yml`, see [Global options](#global-options)), 
-  locally (as described here), or both. 
+  globally (in `mkdocs.yml`, see [Global options](#global-options)),
+  locally (as described here), or both.
 
 !!! example "Example with the Python handler"
     === "docs/my_page.md"
@@ -176,7 +176,7 @@ is possible to link to with `[example][full.path.object1]`, regardless of the cu
 
 ### Cross-references to any Markdown heading
 
-TIP: **Changed in version 0.15.**  
+TIP: **Changed in version 0.15.**
 Linking to any Markdown heading used to be the default, but now opt-in is required.
 
 If you want to link to *any* Markdown heading, not just *mkdocstrings*-inserted items, please
@@ -318,6 +318,18 @@ plugins:
 
 Absolute URLs to cross-referenced items will then be based
 on `https://docs.example.com/version/` instead of `https://cdn.example.com/version/`.
+
+If you need authentication to access the inventory file, you can provide the credentials as environment variables in the URL, either as `username:password`:
+
+```yaml
+- url: https://$MY_USERNAME:$MY_PASSWORD@private.example.com/version/objects.inv
+```
+
+or with token authentication:
+
+```yaml
+- url: https://$TOKEN@private.example.com/version/objects.inv
+```
 
 Reciprocally, *mkdocstrings* also allows to *generate* an inventory file in the Sphinx format.
 It will be enabled by default if the Python handler is used, and generated as `objects.inv` in the final site directory.
--- a/src/mkdocstrings/_cache.py
+++ b/src/mkdocstrings/_cache.py
@@ -4,26 +4,72 @@ import hashlib
 import os
 import urllib.parse
 import urllib.request
-from typing import BinaryIO, Callable
+from typing import BinaryIO, Callable, Union
 
 import click
 import platformdirs
+import regex as re
 
 from mkdocstrings.loggers import get_logger
 
 log = get_logger(__name__)
 
+# Regex pattern for an environment variable
+ENV_VAR_PATTERN = re.compile(r"^\$([A-Za-z_][A-Za-z0-9_]*)$")
+
 
 def download_url_with_gz(url: str) -> bytes:
+    url, auth_header = _extract_auth_from_url(url)
+
     req = urllib.request.Request(  # noqa: S310
         url,
-        headers={"Accept-Encoding": "gzip", "User-Agent": "mkdocstrings/0.15.0"},
+        headers={"Accept-Encoding": "gzip", "User-Agent": "mkdocstrings/0.15.0", **auth_header},
     )
     with urllib.request.urlopen(req) as resp:  # noqa: S310
         content: BinaryIO = resp
         if "gzip" in resp.headers.get("content-encoding", ""):
             content = gzip.GzipFile(fileobj=resp)  # type: ignore[assignment]
         return content.read()
+
+
+def _extract_auth_from_url(url: str) -> tuple[str, dict[str, str]]:
+    """Extracts the username and password from the URL, if present, and returns the URL and the auth header."""
+    parsed_url = urllib.parse.urlparse(url)
+
+    auth_header: dict[str, str] = {}
+    if parsed_url.username:
+        # If the URL contains a username, we assume the user is trying to authenticate
+        auth_header = _create_auth_header(user_env_var=parsed_url.username, pwd_env_var=parsed_url.password)
+        # Remove the auth part from the URL
+        url = parsed_url._replace(netloc=parsed_url.hostname or "").geturl()
+    return url, auth_header
+
+
+def _create_auth_header(user_env_var: str, pwd_env_var: Union[str, None] = None) -> dict[str, str]:
+    """Creates the Authorization header for basic authentication."""
+    user = _get_environment_variable(user_env_var)
+
+    if pwd_env_var is None:
+        # If password is not provided, we assume that the user is using a token
+        log.debug("Using bearer token for authentication")
+        return {"Authorization": f"Bearer {user}"}
+
+    # Else, we assume that the user is using user:password
+    pwd = _get_environment_variable(pwd_env_var)
+    log.debug("Using basic authentication")
+    return {"Authorization": f"Basic {user}:{pwd}"}
+
+
+def _get_environment_variable(env_var: str) -> str:
+    """Extracts the environment variable name from env_var and returns its value."""
+    match = ENV_VAR_PATTERN.match(env_var)
+    if not match:
+        raise ValueError("URL authentication must be specified as environment variables")
+
+    try:
+        return os.environ[match.group(1)]
+    except KeyError as exc:
+        raise ValueError(f"Environment variable '{match.group(1)}' is not set") from exc
 
 
 # This is mostly a copy of https://github.com/mkdocs/mkdocs/blob/master/mkdocs/utils/cache.py
--- a/tests/test_cache.py
+++ b/tests/test_cache.py
@@ -1,0 +1,81 @@
+"""Tests for the internal mkdocstrings _cache module."""
+
+from __future__ import annotations
+
+import pytest
+
+from mkdocstrings import _cache
+
+
+@pytest.mark.parametrize(
+    ("url", "expected_url"),
+    [
+        ("http://host/path", "http://host/path"),
+        ("http://username:password@host/path", "http://host/path"),
+        ("http://token@host/path", "http://host/path"),
+    ],
+)
+def test_extract_auth_from_url(monkeypatch: pytest.MonkeyPatch, url: str, expected_url: str) -> None:
+    """Test extracting the auth part from the URL."""
+    monkeypatch.setattr(_cache, "_create_auth_header", lambda *args, **kwargs: {})
+    result_url, _result_auth_header = _cache._extract_auth_from_url(url)
+    assert result_url == expected_url
+
+
+def test_create_auth_header_basic_auth(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Test creating the Authorization header for basic authentication."""
+    monkeypatch.setenv("MY_USERNAME", "user123")
+    monkeypatch.setenv("MY_PASSWORD", "pass456")
+
+    auth_header = _cache._create_auth_header(user_env_var="$MY_USERNAME", pwd_env_var="$MY_PASSWORD")  # noqa: S106
+    assert auth_header == {"Authorization": "Basic user123:pass456"}
+
+
+def test_create_auth_header_bearer_auth(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Test creating the Authorization header for bearer token authentication."""
+    monkeypatch.setenv("MY_TOKEN", "token123")
+
+    auth_header = _cache._create_auth_header(user_env_var="$MY_TOKEN")
+    assert auth_header == {"Authorization": "Bearer token123"}
+
+
+@pytest.mark.parametrize(
+    ("env_var", "value"),
+    [
+        ("$MY_USERNAME", "some_user"),
+    ],
+)
+def test_get_environment_variable_valid(monkeypatch: pytest.MonkeyPatch, env_var: str, value: str) -> None:
+    """Test getting an environment variable."""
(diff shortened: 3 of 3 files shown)


src/mkdocstrings/_cache.py
```diff
@@ -26,6 +32,46 @@ def download_url_with_gz(url: str) -> bytes:
...
+
+    if pwd_env_var is None:
+        # If password is not provided, we assume that the user is using a token
+        log.debug("Using bearer token for authentication")
+        return {"Authorization": f"Bearer {user}"}
+
+    # Else, we assume that the user is using user:password
+    pwd = _get_environment_variable(pwd_env_var)
+    log.debug("Using basic authentication")
+    return {"Authorization": f"Basic {user}:{pwd}"}
```
pawamoy, line 60: Shouldn't we base64-encode the `user:pwd` part? Or is urllib doing that for us?

src/mkdocstrings/_cache.py
```diff
@@ -4,20 +4,26 @@
...
 import click
 import platformdirs
+import regex as re
 
 from mkdocstrings.loggers import get_logger
 
 log = get_logger(__name__)
 
+# Regex pattern for an environment variable
+ENV_VAR_PATTERN = re.compile(r"^\$([A-Za-z_][A-Za-z0-9_]*)$")
```
pawamoy, line 18: I think I'd prefer directly going with `${...}` syntax, to clearly delimit the variable names. Otherwise I fear it will be asked later and we'll have to support both :sweat_smile: WDYT? It's also [what PDM does](https://pdm-project.org/en/latest/usage/dependency/#hide-credentials-in-the-url), for example.

src/mkdocstrings/_cache.py
```diff
@@ -26,6 +32,46 @@ def download_url_with_gz(url: str) -> bytes:
...
+    pwd = _get_environment_variable(pwd_env_var)
+    log.debug("Using basic authentication")
+    return {"Authorization": f"Basic {user}:{pwd}"}
+
+
+def _get_environment_variable(env_var: str) -> str:
+    """Extracts the environment variable name from env_var and returns its value."""
+    match = ENV_VAR_PATTERN.match(env_var)
+    if not match:
+        raise ValueError("URL authentication must be specified as environment variables")
```
pawamoy, line 67: Do we actually want to prevent users from hardcoding credentials? They're responsible adults.

In other words, couldn't we refactor this to just regex-replace environment variables found in the username and password parts? This way we would directly support `https://prefix${MY_USERNAME}suffix:prefix${MY_PASSWORD}suffix@example.com/path/objects.inv`, which I think could be asked for later anyway.

There's the case where `${VALID_NAME}` appears in a hardcoded username/password though. We probably don't want to output false-positive warnings, but we also want to warn users when they do want to use an env var, but it's unset. Maybe we can just log such warnings/hints when the URL fails to download, a bit later in the process. WDYT? Maybe I'm just over-complicating things too early.

pawamoy (commented):
Thanks a lot! Clean code :+1: A few questions.

src/mkdocstrings/_cache.py
connesy: So only `${...}`, or both? I personally prefer `${...}` as well, but felt the pattern became a bit too complicated when trying to support both:
```python
ENV_VAR_PATTERN = re.compile(r"^\$([A-Za-z_][A-Za-z0-9_]*)$|^\$\{([A-Za-z_][A-Za-z0-9_]*)\}$")
```
plus having to modify the `match.group(1)` to now check and extract from both capture groups.

pawamoy: Only `${...}` :slightly_smiling_face:

connesy: I've changed it to `${...}` in [`3c8a294` (#710)](https://github.com/mkdocstrings/mkdocstrings/pull/710/commits/3c8a2946d15b3c8fd00706d77bb44bdccbfadf2a)

src/mkdocstrings/_cache.py
pawamoy: I can take it from there if you want.

connesy: I'm working on the first part of your suggestion (regex replacing env vars). I found the implementation that [PDM uses](https://github.com/pdm-project/pdm/blob/ea39d366a3d57b46517a1617a96c96a3fccb7d12/src/pdm/utils.py#L252) (thanks to your comment), which seems rather neat.

In regards to your second comment on `${VALID_BUT_UNSET}`, I think it makes sense to log it as a warning. It is probably quite unlikely that e.g. the password happens to match a pattern of `${...}`, but I see no reason to fail hard in that case.

pawamoy: OK, lets go with an early warning :slightly_smiling_face: If people complain we can always move the warning to a later time in the process, as I suggested.

pawamoy: Sure, feel free to add a comment above the relevant functions :slightly_smiling_face:

docs/usage/index.md
```diff
@@ -176,7 +176,7 @@ is possible to link to with `[example][full.path.object1]`, regardless of the cu
 
 ### Cross-references to any Markdown heading
 
-TIP: **Changed in version 0.15.**  
+TIP: **Changed in version 0.15.**
```
pawamoy, line 179:
```suggestion
TIP: **Changed in version 0.15.**  
```

docs/usage/index.md
```diff
@@ -319,6 +319,24 @@ plugins:
 Absolute URLs to cross-referenced items will then be based
 on `https://docs.example.com/version/` instead of `https://cdn.example.com/version/`.
 
+If you need authentication to access the inventory file, you can provide the credentials in the URL, either as `username:password`:
+
+```yaml
+- url: https://username:password@private.example.com/version/objects.inv
+```
+
+or with token authentication:
```
pawamoy, line 328:
```suggestion
...or with token authentication:
```

src/mkdocstrings/_cache.py
```diff
@@ -13,11 +16,16 @@
 
 log = get_logger(__name__)
 
+# Regex pattern for an environment variable in the form ${ENV_VAR}
```
pawamoy, line 19:
```suggestion
# Regex pattern for an environment variable in the form ${ENV_VAR}.
```

src/mkdocstrings/_cache.py
```diff
@@ -26,6 +34,57 @@ def download_url_with_gz(url: str) -> bytes:
...
+        try:
+            return env[match.group(1)]
+        except KeyError:
+            log.warning(f"Environment variable '{match.group(1)}' is not set")
+            return match.group(0)
+
+    return re.sub(ENV_VAR_PATTERN, replace_func, credential)
+
+
+# Implementation adapted from PDM: https://github.com/pdm-project/pdm
```
pawamoy, line 59:
```suggestion
# Implementation adapted from PDM: https://github.com/pdm-project/pdm.
```

src/mkdocstrings/_cache.py
```diff
@@ -26,6 +34,57 @@ def download_url_with_gz(url: str) -> bytes:
...
 
 
+def _expand_env_vars(credential: str, env: Optional[Mapping[str, str]] = None) -> str:
+    """A safe implementation of env var substitution.
+
+    It only supports the following forms:
+
+        ${ENV_VAR}
+
+    Neither $ENV_VAR and %ENV_VAR is supported.
```
pawamoy, line 44:
```suggestion
    It only supports the following forms: `${ENV_VAR}`.
    Neither `$ENV_VAR` or `%ENV_VAR` are supported.
```

src/mkdocstrings/_cache.py
```diff
@@ -26,6 +34,57 @@ def download_url_with_gz(url: str) -> bytes:
         return content.read()
 
 
+def _expand_env_vars(credential: str, env: Optional[Mapping[str, str]] = None) -> str:
+    """A safe implementation of env var substitution.
```
pawamoy, line 38:
```suggestion
    """A safe implementation of environment variable substitution.
```

src/mkdocstrings/_cache.py
```diff
@@ -26,6 +34,57 @@ def download_url_with_gz(url: str) -> bytes:
...
+    auth, host = netloc.split("@", 1)
+    auth = _expand_env_vars(credential=auth)
+    auth_header = _create_auth_header(credential=auth)
+
+    url = urllib.parse.urlunparse((scheme, host, *rest))
+    return url, auth_header
+
+
+def _create_auth_header(credential: str) -> dict[str, str]:
+    """Creates the Authorization header for basic or bearer authentication, depending on credential."""
```
pawamoy, line 75:
```suggestion
    """Create the Authorization header for basic or bearer authentication, depending on credential."""
```

src/mkdocstrings/_cache.py
```diff
@@ -26,6 +34,57 @@ def download_url_with_gz(url: str) -> bytes:
...
+        except KeyError:
+            log.warning(f"Environment variable '{match.group(1)}' is not set")
+            return match.group(0)
+
+    return re.sub(ENV_VAR_PATTERN, replace_func, credential)
+
+
+# Implementation adapted from PDM: https://github.com/pdm-project/pdm
+def _extract_auth_from_url(url: str) -> tuple[str, dict[str, str]]:
+    """Extracts credentials from the URL, if present, and returns the URL and the appropriate auth header for the credentials."""
```
pawamoy, line 61:
```suggestion
    """Extract credentials from the URL if present, and return the URL and the appropriate auth header for the credentials."""
```

src/mkdocstrings/_cache.py
```diff
@@ -26,6 +34,57 @@ def download_url_with_gz(url: str) -> bytes:
...
+    auth_header = _create_auth_header(credential=auth)
+
+    url = urllib.parse.urlunparse((scheme, host, *rest))
+    return url, auth_header
+
+
+def _create_auth_header(credential: str) -> dict[str, str]:
+    """Creates the Authorization header for basic or bearer authentication, depending on credential."""
+    if ":" not in credential:
+        # We assume that the user is using a token
```
pawamoy, line 77:
```suggestion
        # We assume that the user is using a token.
```

src/mkdocstrings/_cache.py
```diff
@@ -26,6 +34,57 @@ def download_url_with_gz(url: str) -> bytes:
...
+        log.debug("Using bearer token for authentication")
+        return {"Authorization": f"Bearer {credential}"}
+
+    # Else, we assume that the user is using user:password
```
pawamoy, line 81:
```suggestion
    # Else, we assume that the user is using user:password.
```

pawamoy (commented):
Thanks for the updates!

Just a round on docstrings and comments. I'll apply in batch. Next I'll suggest a few changes to log messages, to make them more informative.

src/mkdocstrings/_cache.py
```diff
@@ -26,6 +34,54 @@ def download_url_with_gz(url: str) -> bytes:
...
+# Implementation adapted from PDM: https://github.com/pdm-project/pdm.
+def _extract_auth_from_url(url: str) -> tuple[str, dict[str, str]]:
+    """Extract credentials from the URL if present, and return the URL and the appropriate auth header for the credentials."""
+    if "@" not in url:
+        return url, {}
+
+    scheme, netloc, *rest = urllib.parse.urlparse(url)
+    auth, host = netloc.split("@", 1)
+    auth = _expand_env_vars(credential=auth)
+    auth_header = _create_auth_header(credential=auth)
```
pawamoy, line 65:
```suggestion
    auth_header = _create_auth_header(credential=auth, url=url)
```

src/mkdocstrings/_cache.py
```diff
@@ -26,6 +34,54 @@ def download_url_with_gz(url: str) -> bytes:
...
+
+    url = urllib.parse.urlunparse((scheme, host, *rest))
+    return url, auth_header
+
+
+def _create_auth_header(credential: str) -> dict[str, str]:
```
pawamoy, line 71:
```suggestion
def _create_auth_header(credential: str, url: str) -> dict[str, str]:
```

src/mkdocstrings/_cache.py
```diff
@@ -26,6 +34,54 @@ def download_url_with_gz(url: str) -> bytes:
...
+    """Create the Authorization header for basic or bearer authentication, depending on credential."""
+    if ":" not in credential:
+        # We assume that the user is using a token.
+        log.debug("Using bearer token for authentication")
```
pawamoy, line 75:
```suggestion
        log.debug("Using bearer token authentication for %s", url)
```

src/mkdocstrings/_cache.py
```diff
@@ -26,6 +34,54 @@ def download_url_with_gz(url: str) -> bytes:
...
+        return {"Authorization": f"Bearer {credential}"}
+
+    # Else, we assume that the user is using user:password.
+    user, pwd = credential.split(":", 1)
+    log.debug("Using basic authentication")
```
pawamoy, line 80:
```suggestion
    log.debug("Using basic authentication for %s", url)
```

src/mkdocstrings/_cache.py
```diff
@@ -26,6 +34,54 @@ def download_url_with_gz(url: str) -> bytes:
         return content.read()
 
 
+def _expand_env_vars(credential: str, env: Optional[Mapping[str, str]] = None) -> str:
```
pawamoy, line 37:
```suggestion
def _expand_env_vars(credential: str, url: str, env: Optional[Mapping[str, str]] = None) -> str:
```

src/mkdocstrings/_cache.py
```diff
@@ -26,6 +34,54 @@ def download_url_with_gz(url: str) -> bytes:
...
+    Neither `$ENV_VAR` or `%ENV_VAR` are supported.
+    """
+    if env is None:
+        env = os.environ
+
+    def replace_func(match: re.Match) -> str:
+        try:
+            return env[match.group(1)]
+        except KeyError:
+            log.warning(f"Environment variable '{match.group(1)}' is not set")
```
pawamoy, line 50:
```suggestion
            log.warning("Environment variable '%s' is not set, but is used in inventory URL %s", match.group(1), url)
```

pawamoy (commented):
OK WDYT of this? We mention the actual URLs in log messages, to make them more explicit (and/or actionable).

We use `%`-formatting instead of f-strings for performance (and consistency) reasons.

connesy:
> We mention the actual URLs in log messages, to make them more explicit (and/or actionable)

Sure, we can do that.

> We use %-formatting instead of f-strings for performance (and consistency) reasons.

I'm happy to change it to %-formatting for consistency, but as far as I know, f-strings are faster :slightly_smiling_face:

pawamoy:
> I'm happy to change it to %-formatting for consistency, but as far as I know, f-strings are faster

f-strings are faster. But when logging debug messages, if the debug level is not actually enabled, you can skip building the string entirely, which logging supports, but only with %-formatting:

```python
log.debug(f"Debug message with {costly_string_transform}")
log.debug("String transform of %s not performed if logging level is INFO", costly_string_transform)
```

It's generally *not* a perf gain, but for the few times it is, we want the consistency of using %-formatting for every logging message :smile:

connesy:
> f-strings are faster. But when logging debug messages, if the debug level is not actually enabled, you can skip building the string entirely, which logging supports, but only with %-formatting:

Sorry, I didn't notice you were using the logging module to format the message and not `"%s" % value` :slightly_smiling_face:

connesy:
> LGTM! Thank you so much :rocket: 

Thank you for the nice feedback 🙂

PR #748, opened 2025-03-27
fix: Load Sphinx inventories whose dispname spans multiple lines

Yoshanuikabundi:
Hi! I'm Josh, I'm in charge of documentation at [Open Force Field](https://docs.openforcefield.org). We produce a particular kind of mathematical model for computational chemistry. I haven't been using mkdocs/mkdocstrings for long, but I'm enjoying getting to know it, and I'm very excited to write docstrings in markdown!

This is my first contribution here. Thanks for making such clear contributor docs! If you'd rather I do anything another way I'm very happy to fix things. Sorry I haven't raised an issue, but by the time I'd figured out what was going on I basically had the fix written, so it's all here.

## The issue

When a display name in a Sphinx inventory file spans multiple lines, both mkdocstrings and Sphinx simply print the newline character and then the continuation. This is because a display name spanning multiple lines is a rare edge case that was not accounted for in either software. However, while Sphinx discards any inventory lines that do not parse (such as the continuation line), mkdocstrings raises an error, causing the entire inventory file to fail to load.

## Additional context

This continuation line behavior can be seen in the wild in the [OpenEye Toolkits inventory file] and a few Open Force Field [inventory] [files]. These projects' inventories cannot be used with mkdocstrings because of the raised error. You can see this for yourself by uncommenting the relevant lines in OpenFF Pablo's [`mkdocs.yml`], but it might be a pain to collect the dependencies unless you're familiar with Conda (environment file is in `devtools/conda-envs/docs_env.yaml`).

[`mkdocs.yml`]: https://github.com/openforcefield/openff-pablo/blob/177a0e94bbbb74d90e3a80ac08cad998e6b3ac15/docs/mkdocs.yml#L139-L148
[OpenEye Toolkits inventory file]: https://docs.eyesopen.com/toolkits/python/objects.inv
[inventory]:  https://docs.openforcefield.org/projects/toolkit/en/stable/objects.inv
[files]: https://docs.openforcefield.org/projects/bespokefit/en/stable/objects.inv

Note that in Sphinx, these inventory files are read successfully, but the continuation lines are discarded, truncating the display name. In theory, a continuation line that by chance did parse correctly would be interpreted by both packages as a new inventory item.

## Changes in this PR

This PR appends the continuation line to the previous inventory item display name when it fails to parse. This allows these inventory files to be loaded while preserving the original display name. Note that the theoretical continuation line that parses by chance is not addressed. Round-trip tests and additional Sphinx tests are added to check the new behavior. I've run these tests locally with Sphinx added to the dependencies and they do actually succeed, and I've also confirmed that they fail without the new code to confirm the tests are doing something.

I also found that `make format` and `make check` introduced several formatting changes and new lint failures outside the files I changed; I determined that they were due to the recent new version of Ruff and corrected them in a separate commit within this PR. One judgement I made there that deserves particular review is that a [new Ruff lint] was triggered, apparently over-eagerly; applying the suggested fix where it made sense caused the tests to fail. I therefore set the lint to be ignored. There are some more details in the commit message. Happy to tweak this (and everything else), but it seemed like the existing code was using `markupsafe.Markup` as intended to declare that a certain bit of dynamically generated markup was correct, and the lint was applying too strict a standard.

[new Ruff lint]: https://docs.astral.sh/ruff/rules/unsafe-markup-use/

## Other stuff

Thanks for all your work on mkdocstrings!

### DIFF
--- a/src/mkdocstrings/_internal/inventory.py
+++ b/src/mkdocstrings/_internal/inventory.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 import re
 import zlib
 from textwrap import dedent
-from typing import TYPE_CHECKING, BinaryIO
+from typing import TYPE_CHECKING, BinaryIO, Literal, overload
 
 if TYPE_CHECKING:
     from collections.abc import Collection
@@ -66,11 +66,21 @@ class InventoryItem:
     sphinx_item_regex = re.compile(r"^(.+?)\s+(\S+):(\S+)\s+(-?\d+)\s+(\S+)\s*(.*)$")
     """Regex to parse a Sphinx v2 inventory line."""
 
+    @overload
     @classmethod
-    def parse_sphinx(cls, line: str) -> InventoryItem:
+    def parse_sphinx(cls, line: str, *, return_none: Literal[False]) -> InventoryItem: ...
+
+    @overload
+    @classmethod
+    def parse_sphinx(cls, line: str, *, return_none: Literal[True]) -> InventoryItem | None: ...
+
+    @classmethod
+    def parse_sphinx(cls, line: str, *, return_none: bool = False) -> InventoryItem | None:
         """Parse a line from a Sphinx v2 inventory file and return an `InventoryItem` from it."""
         match = cls.sphinx_item_regex.search(line)
         if not match:
+            if return_none:
+                return None
             raise ValueError(line)
         name, domain, role, priority, uri, dispname = match.groups()
         if uri.endswith("$"):
@@ -167,7 +177,9 @@ class Inventory(dict):
         for _ in range(4):
             in_file.readline()
         lines = zlib.decompress(in_file.read()).splitlines()
-        items = [InventoryItem.parse_sphinx(line.decode("utf8")) for line in lines]
+        items: list[InventoryItem] = [
+            item for line in lines if (item := InventoryItem.parse_sphinx(line.decode("utf8"), return_none=True))
+        ]
         if domain_filter:
             items = [item for item in items if item.domain in domain_filter]
         return cls(items)
--- a/config/ruff.toml
+++ b/config/ruff.toml
@@ -43,6 +43,7 @@ ignore = [
     "PLR0913",  # Too many arguments to function call
     "PLR0915",  # Too many statements
     "SLF001", # Private member accessed
+    "S704", # Unsafe use of `markupsafe.Markup`
     "TRY003",  # Avoid specifying long messages outside the exception class
 ]
 
--- a/scripts/insiders.py
+++ b/scripts/insiders.py
@@ -168,6 +168,6 @@ goals = funding_goals(data_source, funding=current_funding)
 ongoing_goals = [goal for goal in goals.values() if not goal.complete]
 unreleased_features = sorted(
     (ft for ft in feature_list(ongoing_goals) if ft.since),
-    key=lambda ft: cast(date, ft.since),
+    key=lambda ft: cast("date", ft.since),
     reverse=True,
 )
--- a/src/mkdocstrings/_internal/handlers/base.py
+++ b/src/mkdocstrings/_internal/handlers/base.py
@@ -17,7 +17,6 @@ from xml.etree.ElementTree import Element, tostring
 
 from jinja2 import Environment, FileSystemLoader
 from markdown import Markdown
-from markdown.extensions.toc import TocTreeprocessor
 from markupsafe import Markup
 from mkdocs.utils.cache import download_and_cache_url
 from mkdocs_autorefs import AutorefsInlineProcessor, BacklinksTreeProcessor
@@ -33,7 +32,6 @@ from mkdocstrings._internal.handlers.rendering import (
 from mkdocstrings._internal.inventory import Inventory
 from mkdocstrings._internal.loggers import get_logger, get_template_logger
 
-
 # YORE: EOL 3.9: Replace block with line 4.
 if sys.version_info < (3, 10):
     from importlib_metadata import entry_points
@@ -44,6 +42,7 @@ if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator, Mapping, Sequence
 
     from markdown import Extension
+    from markdown.extensions.toc import TocTreeprocessor
     from mkdocs_autorefs import AutorefsHookInterface, Backlink
 
 _logger = get_logger(__name__)
@@ -494,7 +493,7 @@ class BaseHandler:
         el = Element(f"h{heading_level}", attributes)
         el.append(Element("mkdocstrings-placeholder"))
         # Tell the inner 'toc' extension to make its additions if configured so.
-        toc = cast(TocTreeprocessor, self.md.treeprocessors["toc"])
+        toc = cast("TocTreeprocessor", self.md.treeprocessors["toc"])
         if toc.use_anchors:
             toc.add_anchor(el, attributes["id"])
         if toc.use_permalinks:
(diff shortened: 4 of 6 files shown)


src/mkdocstrings/_internal/inventory.py
```diff
@@ -167,7 +167,20 @@ def parse_sphinx(cls, in_file: BinaryIO, *, domain_filter: Collection[str] = ())
         for _ in range(4):
             in_file.readline()
         lines = zlib.decompress(in_file.read()).splitlines()
-        items = [InventoryItem.parse_sphinx(line.decode("utf8")) for line in lines]
+        items: list[InventoryItem] = []
+        for line in lines:
+            if len(line) == 0 and len(items) == 0:
+                # Skip empty lines at start of inventory section
+                continue
```
Yoshanuikabundi, line 174: This bit just allows inventory files without any items to be loaded, which is used in the round trip test; it can be removed without any dramas, but the empty inventory will also need to be removed from the round trip test (but not the Sphinx loading test).

pawamoy:
Hi @Yoshanuikabundi, thanks for PR and detailed explanation!

Looking at [Sphobjinv's docs](https://sphobjinv.readthedocs.io/en/latest/syntax.html) (from @bskinn):

> {dispname}
> Default cross-reference text to be displayed in compiled documentation.
> Constraints:
> - MUST have nonzero length
> - MAY contain internal whitespace (leading/trailing whitespace is ignored)

...it seems newlines are indeed valid inside `dispname`. @bskinn could you confirm? :pray: 

Out of curiosity @Yoshanuikabundi, could you show some example `disname` values that are present in the mentioned inventories?

bskinn:
> ...it seems newlines are indeed valid inside `dispname`.

My first reaction: I _believe_ this is >incorrect<, and in fact newlines are _invalid_ anywhere in any field in a v2 `objects.inv` file.

I would have to dig back into the Sphinx code to be sure, but I'm pretty sure that Sphinx processes the decoded `objects.inv` payload line by line, always treating each line separately as a candidate inventory item. Any line that doesn't parse as a valid inventory object is silently ignored, leading to the behavior noted by OP:

> However, while Sphinx discards any inventory lines that do not parse (such as the continuation line) ...

I wrote all of of the higher-level components of the `sphobjinv` API to mimic this behavior, for maximum compatibility. I don't have a ticket for it at the moment, but a good feature for `sphobjinv` (and for Sphinx too, for that matter) would be to emit a warning if any lines in an inventory were ignored due to syntax mismatch. (No ticket at the moment on `sphobjinv`; I opened https://github.com/sphinx-doc/sphinx/issues/8825 some time ago, but neither I nor anyone else has worked on it, AFAIK.)

My opinion here is that `mkdocstrings` revise to match Sphinx's behavior, and not add handling for newlines in `dispname`.

This seems to me to be an [XY problem](https://en.wikipedia.org/wiki/XY_problem). IMO, `dispname`s should be _short_, at most a small phrase; and, thus any text substantial enough to need a newline is not appropriate for a `dispname`, whether or not the `dispname` machinery happens to work for the desired use-case. In the specific examples I found below, the majority appear to be figure captions -- perhaps the intention for these is to create a Table of Figures? If so, I would think these newlines within the `dispnames` would be problematic if included in that ToF.

In any event, I would guess that the best solution here to the documentation need for these projects is a Sphinx extension of some kind.

---

I found the mismatching objects in the OpenEye Toolkits `objects.inv` with the following:

```py
>>> import sphobjinv as soi
>>> from pathlib import Path
>>> cbdata = Path('objects.inv').read_bytes()
>>> bdata = soi.decompress(cbdata)
>>> tdata = bdata.decode()
>>> [l for l in tdata.splitlines()[4:] if not soi.re.p_data.match(l)]  # [4:] skips the header

... list of syntax-mismatching lines ...

```

I then exported the decompressed contents to disk with:

```py
>>> Path('objects-manual.txt').write_text(tdata)
```

And searched for the mismatching lines. Here are some examples of full entries with internal newlines from that same `objects.inv`:

```none
figure_fragmentation std:label -1 medchemtk/fragmentation.html#figure-fragmentation Example of fragmentation (A) input molecule
(B) fragments returned by the OEGetRingChainFragments function

figure_oe2datomdisplay_getpropertyoffset std:label -1 depicttk/OEDepictClasses/OE2DAtomDisplay.html#figure-oe2datomdisplay-getpropertyoffset The property offset vector is relative to coordinates returned by
the OE2DAtomDisplay::GetCoords method

figure_oe2dbonddisplay_getpropertyoffset std:label -1 depicttk/OEDepictClasses/OE2DBondDisplay.html#figure-oe2dbonddisplay-getpropertyoffset The property offset vector is relative to the middle of the
displayed bond

figure_ring_sssr std:label -1 oechemtk/ring.html#figure-ring-sssr Example of SSSR
SSSR is not an invariant subset of all possible rings; (a), (b) and (c) depict the three equally valid SSSR of the bridged structure (on the left)

posit_cross_docking_results std:label -1 dockingtk/theory/shapefit.html#posit-cross-docking-results SHAPEFIT Cross Docking Results: Probability of finding a good pose based on bound-ligand
fit-ligand TanimotoCombo similarity.  Standard docking results
are essentially the same and follow the same trajectories
flattening out as they hit their limit of accuracy.  While SHAPEFIT
performs worse at low similarities it continually increases as
similarity increases.

posit_heat_map std:label -1 dockingtk/theory/pose_stub_tk.html#posit-heat-map POSIT Probability Map: Given a 2D similarity (in this case the MACCS 166 descriptor set [Durant-2002])
and a 3D similarity (TanimotoCombo) posit computes a
probability of finding the correct pose based on an analysis of
historical and experimental data.

```

bskinn:
One thing this discussion _does_ do for `sphobjinv`, is highlight that additional specificity is needed on that docs page. 

The invalidity of newlines in any field is implicit in the use of "data line" in this sentence on that page (emphasis added):

> The descriptions below have been updated to reflect this and to provide more detailed information on the constraints governing each field of ***an `objects.inv` data line***.

But the description could use a more direct statement of the strict limitation of each data item being constrained to a single line of the decompressed inventory source.

pawamoy:
Thank you @bskinn!

My initial reaction was similar: even if newlines were valid, how they are rendered depend on the output format (HTML, Markdown, PDF, etc.) and so I wasn't sure it made sense to support them.

Thanks a lot for searching for and finding examples in the mentioned inventories. These "display names" definitely look like *descriptions*, not *names*, so I'd be inclined to say they are invalid.

In short, I agree with @bskinn's reasoning and suggestions, and will probably not merge this PR as it is. Instead we should log warnings on invalid lines.

pawamoy:
In:

https://github.com/sphinx-doc/sphinx/blob/master/sphinx/util/inventory.py#L120-L121

...non-matching lines are indeed skipped.

Yoshanuikabundi:
That sounds very reasonable to me! The offending lines are mostly automatically generated from figure/image captions and are very unlikely to be desired in any cross reference. I'll revise this PR to simply discard lines that can't be parsed rather than append them to the previous entry, but I won't be able to get to this immediately.

I'd be against issuing a warning - the warning would usually not be actionable by the users as by definition it comes from another documentation site. In fact, in this case even the docs that the inventory was from wouldn't be able to action the warning as its more or less a misjudgement in a Sphinx plugin. I'd worry it'd almost always just be noise, which mkdocs is usually good about minimising. I don't believe Sphinx warns.

pawamoy:
Ah, you're completely right @Yoshanuikabundi, good point. No warning log then! And no worries, I'll take it from here if you don't have the time :slightly_smiling_face:

Yoshanuikabundi:
Great, thanks @pawamoy!

pawamoy (approved):

PR #765, opened 2025-05-14
Increase minimum version of python-legacy handler

mschoettle:
Just updating `mkdocstrings` without raising the `mkdocstrings-python-legacy` causes lots of deprecation warnings when the latter is still at `0.2.4`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ dependencies = [
 
 [project.optional-dependencies]
 crystal = ["mkdocstrings-crystal>=0.3.4"]
-python-legacy = ["mkdocstrings-python-legacy>=0.2.1"]
+python-legacy = ["mkdocstrings-python-legacy>=0.2.5"]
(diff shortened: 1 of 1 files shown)


mschoettle:
It looks like the job failure is caused by something unrelated:

>   ERROR   -  Config value 'plugins': Plugin 'llmstxt' option 'sections': Required configuration not provided.

pawamoy:
Thanks for the PR @mschoettle!

However deprecation warnings are just warnings, and shouldn't require that we force users to update mkdocstrings-python-legacy. I'm not sure in which case you'd want to update mkdocstrings but not mkdocstrings-python-legacy, but users are still allowed to do that.

WDYT?

pawamoy:
Thanks for the explanation @mschoettle. In this case I'll close as I don't think we should force the update here. Feel free to open another issue for the one you're having :slightly_smiling_face:

